PR #45, opened 2022-01-26
Input clash handling

### LINKED ISSUE #23: Add tools to better handle overlapping inputs
E.g. have both `Ctrl+R` and `R` bound, but only trigger want to trigger `Ctrl+R` hotkey.

Perhaps trigger on release in case of conflict?

Specify dedicated modifier keys?

alice-i-cecile:
## Problem

Fixes #23.

In many applications, overlapping inputs are special cased, to avoid surprising results, especially with modifier keys.

If `Ctrl + V` and `Ctrl + Shift + V` only pastes without formatting, rather than doing that *and* pasting normally.

## Solution

Two inputs are said to clash if one is a strict subset of the other.

Allow users to configure their input maps to decide how they want to handle clashing inputs.

```rust
pub enum ClashStrategy {
    /// All matching inputs will always be pressed
    PressAll,
    /// Only press the action that corresponds to the longest chord
    PrioritizeLongest,
    /// If the [`UserInput`] contains a modifier key (defined at the input map level), press that action over any unmodified action.
    ///
    /// If more than one matching action uses a modifier, break ties based on number of modifiers.
    /// Further ties are broken using the `PrioritizeLongest` rule.
    PrioritizeModified,
    /// Use the order in which actions are defined in the enum to resolve clashing inputs
    ///
    /// Uses the iteration order returned by [IntoEnumIterator](crate::IntoEnumIterator),
    /// which is generated in order of the enum items by the `#[derive(EnumIter)]` macro.
    UseActionOrder,
}
```

## To do

- [x] actually handle clashes according to the selected strategy
- [x] add integration tests to verify that this works as expected
- [x] add an example demonstrating how this works

### DIFF
--- a/.github/linters/.markdown-link.yml
+++ b/.github/linters/.markdown-link.yml
@@ -1,4 +1,0 @@
-{
-    "MD013": false
-    "MD024:" false
-}
--- a/.github/linters/.markdown-link.yml
+++ b/.github/linters/.markdown-link.yml
@@ -1,3 +1,4 @@
 {
     "MD013": false
+    "MD024:" false
 }
--- a/.github/linters/.markdown-lint.yml
+++ b/.github/linters/.markdown-lint.yml
@@ -12,3 +12,4 @@
 # Rules by tags #
 #################
 no-duplicate-heading: false
+line-length: false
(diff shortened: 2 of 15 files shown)


alice-i-cecile:
Done! Merging now.

Issue #19, opened 2021-12-27
Add serialization and deserialization for InputMap and ActionState


alice-i-cecile:
## What problem does this solve?

Save and load my input map settings to a file.

## What solution would you like?

Implement `Serialize` and `Deserialize` for the `InputMap`, `ActionState` and `ActionDiff` types.

alice-i-cecile:
Blocked by serialization impls in `PetitSet` and on `Gamepad`.

alice-i-cecile:
Actually, blocked on serialization impls for all of the button types and `PetitSet`.

alice-i-cecile:
Blocked on https://github.com/bevyengine/bevy/issues/3794

PR #61, opened 2022-01-31
ActionDiff: a network efficient serialization struct

### LINKED ISSUE #58: Add action event stream type
This is extremely useful for serialization, allowing users to send the bare minimum of state across the network.

Outside of these use cases, these is generally just less clear and reliable than examining the actionstate itself, so it should be split off into its own plugins: one for reading, and another for writing.

alice-i-cecile:
Fixes #58.

Check out the new `send_actions_over_network` example for usage.


alice-i-cecile:
Next up: #19.

Issue #60, opened 2022-01-30
Store action-specific state in `ActionState`


alice-i-cecile:
Many actions are only meaningful in association with non-enumerable data.

A very useful pattern here is to store information in the enum variant, and then update that appropriately.

We want to be able to express:
```rust
#[derive(Actionlike, Clone)]
enum ArpgAction {
    Movement(Direction),
    Ability1,
    Ability2,
    Ability3,
    Ability4,
    Ultimate,
}

input_map.insert(ArpgAction::Movement(Direction::NORTH), KeyCode::Up);

let ArpgAction::Movement(direction) = action_state.get_value::<ArpgAction::Movement>();
```

Unfortunately, a reasonable API for the last step seems to be completely blocked on https://github.com/rust-lang/rfcs/pull/2593, which, unsurprisingly, is postponed indefinitely 🙃

alice-i-cecile:
Blocks #50.

alice-i-cecile:
We can eliminate the `Hash` bound by storing a `HashMap<usize, InputAction>` in `InputMap`, and doing likewise for `ActionMap`. This usize is given by `Actionlike::index()`, which has the desired behavior!

This may make #68 harder, but this feature is significantly more important, and serves related goals.

alice-i-cecile:
We may be able to workaround this by generating an internal type to match those of each variant, and then requesting those in `ActionState::get_value` :/ It's *extremely* cursed though due to the high levels of magic, and will do wonky things with the imports.

alice-i-cecile:
This is fundamentally wrong-headed, and not our responsibility.

alice-i-cecile:
#151 has some related mechanisms, we should figure out how to steal from the approach used to store gamepad axis values.

alice-i-cecile:
Closing this out because we can now store action state values.

bardt:
@alice-i-cecile could you (or someone else) point to an example of how to store action state value? Or at least to the feature you mean.

I want to do something similar to the example in the issue description: pass additional data with an Actionlike enum variant.

alice-i-cecile:
The feature that I've been using is the `DualAxis` / `SingleAxis` features, where we can store 2/1 f32s. Is that enough for your use case?

I'm not fully satisfied with the design here either/

bardt:
[...]

PR #67, opened 2022-02-09
Release 0.3

### LINKED ISSUE #73: Resolve button conflicts
Let's assume we have input attached to two entities, one has mapped a chord [ctrl, a], and the other just [a]. If ctrl-a is pressed, the common sense response would be only to run the first action, but in this plugin both are run. 

The problem seems to be already solved in alternative crates, e.g. https://github.com/lightsoutgames/bevy_input_actionmap, would be nice to have it here too.

alice-i-cecile:
This release is planned to center around:

- ~~handling game stick inputs~~
- assorted usability improvements
- clash handling

## Issues

- Closes #73

## Critical questions

- [x] if we gather axis data into an `AxisPair` struct, where is it stored?
  - Store as `Res<AxisPairs<Gamepad>>`, with the ability to extend to other simulated axes
- [x] how are axis-like inputs used to set `ActionState`?
   - new `UserInput` variant based on direction partition clamping
   - set value of variant based on AxisPair direction / vec2 / rotation
- [x] if we want to store data in actionlike variants, how do we generate a variant-only Hash impl?
  - This isn't feasible: see #60. We also no longer need `Hash` 😄
- [x] do we care, at all, about non-enum Actionlike types? If yes, how do we refactor to support this?
  - Nope! See below.


chrisburnor:
>  do we care, at all, about non-enum Actionlike types? If yes, how do we refactor to support this?
 
In my personal input manager, I implemented a system that used `HashMap<TypeId, KeyPress>` to allow heterogeneous types of actions. I also successfully experimented with using PolyMaps and TypeMaps to allow metadata to be attached to the actions rather than just using the TypeId. So, those approaches could work for the refactor.

However, after doing all that, I'm not convinced that non-enum actions are actually worth the overhead. Most games are going to end up with a relatively finite number of 'moves' corresponding to the finite gamepad keys and if one wants to let players remap the keys and then save that option, they're going to need to implement De/Serialize.

alice-i-cecile:
@chrisburnor thanks for the feedback!

> However, after doing all that, I'm not convinced that non-enum actions are actually worth the overhead

Yeah, this is about my conclusion too: it seems like such a rare and unintuitive use case that it's not worth the significant complexity to write and maintain.

chrisburnor:
> in order to store arpeggios and grace periods, we must store a history of relevant input events. Where and how should this be done?

What sort of requirements does this need? Something like a circular queue implemented as an index pointer and fix-length Vec (or slice??) would allow for very fast reading/writing with no memory allocations or resizing, but I don't know if that is actually what we need here. Does it need to be thread safe? Do we need one per 'player'/input-able object? or can one global suffice?

alice-i-cecile:
@chrisburnor, these are fantastic questions, and I'm genuinely not sure yet. As a result, I've removed arpeggios from the release, and plan to implement them later to avoid blocking other features.

PR #82, opened 2022-03-19
Move bevy_ui dependency under a feature

### LINKED ISSUE #80: Turn input_mocking into a feature

Currently `leafwing_input_manager` depends on `bevy_ui` because of the `input_mocking` module.

I see several options:

1. Turn `input_mocking` into a feature.
2. Move `input_mocking` into a separate plugin.
3. Add a feature that disables `bevy_ui` dependency for `input_mocking`

Currently used `render` feature is a [group](
https://github.com/bevyengine/bevy/blob/a291b5aaedb4affcb31df2e2e63cb0c665ffb24a/Cargo.toml#L35-L43) that contains unneeded stuff, like `bevy_ui` or `bevy_text`. I was about to submit a PR but noticed that `bevy_ui` is a required dependency for this plugin.

Shatur:
I moved all `bevy_ui` related functionality under a feature, closes #80.

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -125,6 +125,7 @@ fn run_in_state() {
 }
 
 #[test]
+#[cfg(feature = "ui")]
 fn action_state_driver() {
     use bevy::input::InputPlugin;
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,8 +17,12 @@ opt-level = 3
 [workspace]
 members = ["./", "tools/ci", "macros"]
 
+[features]
+default = ['ui']
+ui = ['bevy/bevy_ui']
+
 [dependencies]
-bevy = {version = "0.6", default-features = false, features = ["render", "serialize"]}
+bevy = {version = "0.6", default-features = false, features = ["serialize"]}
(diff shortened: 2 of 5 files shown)


alice-i-cecile:
Can you retarget this to the `dev` branch? Unfortunately I can't use `main` for active development without eternally confusing users about broken examples.

Shatur:
Oh, sorry, totally forgot about it. Let me rebase the changes.

Shatur:
On `dev` branch you have one doc-test broken before my changes.

alice-i-cecile:
Yep; I was aware that `dev` is currently failing. Thanks for the PR!

tests/integration.rs
```diff
@@ -125,6 +125,7 @@ fn run_in_state() {
 }
 
 #[test]
+#[cfg(feature = "ui")]
```
alice-i-cecile, line 128: Can you please add `bevy_ui` as a dev dependency and remove the feature flag here? Testing in various configurations is generally a pain.

Shatur: But `ui` is enabled by default. So you will be able to run all tests without changing the configuration.

PR #87, opened 2022-03-29
Simplify InputMap

Shatur:
In this PR I make `InputMap` API more map-like. Also now it panics when something is going wrong instead of silently fails.

Should I replace `get` to return an iterator instead of `PetitiSet` directly?

### DIFF
--- a/src/input_map.rs
+++ b/src/input_map.rs
@@ -165,6 +165,7 @@ impl<A: Actionlike> InputMap<A> {
     /// Insert a mapping between `action` and `input`
     ///
     /// # Panics
+    ///
     /// Panics if the map is full and `input` is not a duplicate or `input` is [`UserInput::Null`].
     pub fn insert(&mut self, action: A, input: impl Into<UserInput>) -> &mut Self {
         let input = input.into();
@@ -187,6 +188,7 @@ impl<A: Actionlike> InputMap<A> {
     /// If a matching input already existed in the set, it will be moved to the supplied index. Any input that was previously there will be moved to the matching input’s original index.
     ///
     /// # Panics
+    ///
     /// Panics if the map is full and `input` is not a duplicate or `input` is [`UserInput::Null`].
     pub fn insert_at(&mut self, action: A, input: impl Into<UserInput>, index: usize) -> &mut Self {
         let input = input.into();
@@ -211,6 +213,7 @@ impl<A: Actionlike> InputMap<A> {
     /// Any iterator that can be converted into a [`UserInput`] can be supplied.
     ///
     /// # Panics
+    ///
     /// Panics if the map is full and any of `inputs` is not a duplicate or any of `inputs` is [`UserInput::Null`].
     pub fn insert_multiple(
         &mut self,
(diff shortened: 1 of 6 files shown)


alice-i-cecile:
> Should I replace get to return an iterator instead of PetitiSet directly?

No, I think that returning the `PetitSet` is more helpful to a more diverse collection of end users.

src/input_map.rs
```diff
@@ -166,31 +164,39 @@ impl<A: Actionlike> InputMap<A> {
...
+    ///
+    /// If a matching input already existed in the set, it will be moved to the supplied index. Any input that was previously there will be moved to the matching input’s original index.
+    ///
+    /// # Panics
+    /// Panics if the map is full and `input` is not a duplicate or `input` is [`UserInput::Null`].
+    pub fn insert_at(&mut self, action: A, input: impl Into<UserInput>, index: usize) -> &mut Self {
+        let input = input.into();
+
+        // Don't insert Null inputs into the map
+        if input == UserInput::Null {
```
alice-i-cecile, line 195: Can you please remove UserInput::Null in this PR? @sixfold correctly pointed out that this wasn't very useful.

src/input_map.rs
```diff
@@ -221,7 +228,8 @@ impl<A: Actionlike> InputMap<A> {
     /// Any iterator that can be converted into a [`Button`] can be supplied, but will be converted into a [`PetitSet`] for storage and use.
     /// Chords can also be added with the [insert](Self::insert) method, if the [`UserInput::Chord`] variant is constructed explicitly.
     ///
-    /// Existing mappings for that action will not be overwritten.
+    /// # Panics
```
alice-i-cecile, line 231:
```suggestion
    /// # Panics
    ///
```

src/input_map.rs
Shatur: You mean to remote it at all?

src/input_map.rs
```diff
@@ -429,57 +348,18 @@ impl<A: Actionlike> InputMap<A> {
...
-
-    /// Returns how many bindings are currently registered for the provided action with the provided [`InputMode`]
-    ///
-    /// If `None` is provided, a total across all input modes will be provided.
-    ///
-    /// A maximum of `CAP` bindings across all input modes can be stored for each action,
-    /// and insert operations will silently fail if used when `CAP` bindings already exist.
+    /// Returns the `action` mappings
     #[must_use]
-    pub fn n_registered(&self, action: A, input_mode: Option<InputMode>) -> usize {
```
alice-i-cecile, line 471: This API is still useful, especially to avoid unexpected panics.

src/input_map.rs
Shatur: If yes, how would you like to handle chords coertion if there is no inputs?

src/input_map.rs
Shatur: But maybe just use `get(action).len()`?

src/input_map.rs
alice-i-cecile: I meant remove it entirely, yes.

Which chord coercion in particular were you wondering about?

src/input_map.rs
alice-i-cecile: Yeah okay fair :)

src/input_map.rs
Shatur: I mean how to handle 0 in this case:

https://github.com/Leafwing-Studios/leafwing_input_manager/blob/2496c38d756275b7acedb19a5fc62b57a2632fbd/src/buttonlike_user_input.rs#L52-L56

Just create an empty chord?

alice-i-cecile: Yeah, I think we should create an n-length chord in all cases there, rather than trying to be clever.

PR #93, opened 2022-03-31
Serde tests

Shatur:
I added simple tests just to quickly test serde.
For real tests we should use [test_serde](https://serde.rs/unit-testing.html).


Shatur:
We can serialize into toml because it doesn't support enums serialization into (only deserialization is supported).

Issue #86, opened 2022-03-29
ActionState::release doesn't work as expected


Shatur:
## Version

6b89a11f4550634fd23ad431c426a57dde0c3e66 (dev branch)

## Operating system & version

ArchLinux

## What you did

The following code:

```rust
use bevy::prelude::*;
use leafwing_input_manager::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_plugin(InputManagerPlugin::<Action>::default())
        .add_startup_system(spawn_player)
        .add_system(jump.label("jump"))
        .add_system(consume.before("jump"))
        .run();
}

#[derive(Actionlike, PartialEq, Eq, Clone, Copy, Hash, Debug)]
enum Action {
    Jump,
}

#[derive(Component)]
struct Player;

fn spawn_player(mut commands: Commands) {
    commands
        .spawn()
        .insert(Player)
        .insert_bundle(InputManagerBundle::<Action> {
            action_state: ActionState::default(),
            input_map: InputMap::new([(Action::Jump, KeyCode::Space)]),
        });
}

fn jump(query: Query<&ActionState<Action>, With<Player>>) {
    let action_state = query.single();
    if action_state.just_pressed(Action::Jump) {
        println!("I'm jumping!");
    }
}


fn consume(mut query: Query<&mut ActionState<Action>, With<Player>>) {
    let mut action_state = query.single_mut();
    if action_state.just_pressed(Action::Jump) {
        println!("Not allowing to jump, consuming, should be called one, but...");
        action_state.release(Action::Jump);

    }
}
```

## What you expected to happen

Message "Not allowing to jump, consuming, should be called one, but..." prints once because `consume` releases the button. But on next frame the button becomes pressed again and `consume` releases it again... 

## What actually happened

Message "Not allowing to jump, consuming, should be called one, but..." prints multiple times.

alice-i-cecile:
Fixed by #97

alice-i-cecile:
Still a concern.

PR #97, opened 2022-04-01
Changed `ActionState.button_states` inserts into direct assignments

elfein727:
Before this change, `.insert(...)` on `ActionState::button_states` would duplicate items. Changed those lines to direct assignments so duplication doesn't occur.


Shatur (approved):
Works as expected on my machine, fixes #86.

PR #98, opened 2022-04-02
Iterate over actions with mapping by default

Shatur:
In previous PR I added a method to iterate over inputs. But I realized that such iteration should be with action like in normal map.

So I renamed the submitted `iter()` method into `iter_inputs()` and added `iter()` that iterates over actions with inputs.

I also have to delete `IntoIter` currently because we can't use opaque type in traits without nightly compiler.
If you know how to implement it properly - I would appreciate your advice. But for now I removed it.


alice-i-cecile:
> If you know how to implement it properly - I would appreciate your advice. But for now I removed it.

Not immediately sure; I think this is the right call for now.

PR #106, opened 2022-04-07
Replace run_in_state with DisableInput resource

### LINKED ISSUE #105: Replace run_in_state with resource for more flexibility

Currently we have `run_in_state` function to run the plugin only in specific state. But the approach is quite limiting because in games you often need to dynamically disable the input (for example, when you open menu or chat) and Bevy is moving to the stageless approach which will require users to split UI and game states. 

I would suggest to introduce a resource which allows to disable input if present.

We could query the optional component here:
https://github.com/Leafwing-Studios/leafwing_input_manager/blob/1e95c4c4a8d7041a51b83ec6bd89a3146694e5f1/src/systems.rs#L40
And do not update `ActionState` if it present. Also we should add a separate system which clears `ActionState` on `DisableInput` insertion.

Shatur:
Closes #105.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -17,7 +17,7 @@
 - relaxed `Hash` and `Eq` bounds on `Actionlike`
 - `ActionState::state` and `set_state` methods renamed to `button_state` and `set_button_state` for clarity
 - removed `UserInput::Null`.
-- `InputManagerPlugin::run_in_state` was replaced with `InputDisabled<Actionlike>` resource. Insert when you want to suppress input collection.
+- `InputManagerPlugin::run_in_state` was replaced with `InputDisabled<A: Actionlike>` resource. Insert this resource when you want to suppress input collection, and remove it when you're done.
 - refactored `InputMap`.
   - removed methods that works with specific input mode.
   - removed `n_registered`, use `get(action).len()` instead.
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -17,6 +17,7 @@
 - relaxed `Hash` and `Eq` bounds on `Actionlike`
 - `ActionState::state` and `set_state` methods renamed to `button_state` and `set_button_state` for clarity
 - removed `UserInput::Null`.
+- `InputManagerPlugin::run_in_state` was replaced with `InputDisabled<Actionlike>` resource. Insert when you want to suppress input collection.
 - refactored `InputMap`.
   - removed methods that works with specific input mode.
   - removed `n_registered`, use `get(action).len()` instead.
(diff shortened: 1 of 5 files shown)


alice-i-cecile:
Can you please add this to the Release Notes? This will require migration guidance.

Shatur:
Okay! I also forget to apply formatting.

Also maybe we should insert systems right away instead of putting them into `input_manager_systems` system set?

src/systems.rs
```diff
@@ -140,3 +169,17 @@ pub fn process_action_diffs<A: Actionlike, ID: Eq + Component + Clone>(
         }
     }
 }
+
+/// Release all inputs if [`DisableInput`] was added
```
alice-i-cecile, line 173:
```suggestion
/// Release all actions in all [`ActionState`]s if [`DisableInput`] was added
```
It's important to be clear about the distinction between inputs and actions here.

alice-i-cecile:
> Also maybe we should insert systems right away instead of putting them into input_manager_systems system set?

Agreed, we can do this easily :)

Shatur:
> Agreed, we can do this easily now :)

Will do as a part of a separate PR for easy review.

RELEASES.md
```diff
@@ -17,6 +17,7 @@
 - relaxed `Hash` and `Eq` bounds on `Actionlike`
 - `ActionState::state` and `set_state` methods renamed to `button_state` and `set_button_state` for clarity
 - removed `UserInput::Null`.
+- `InputManagerPlugin::run_in_state` was replaced with `InputDisabled<Actionlike>` resource. Insert when you want to suppress input collection.
```
alice-i-cecile, line 20:
```suggestion
- `InputManagerPlugin::run_in_state` was replaced with `InputDisabled<A: Actionlike>` resource. Insert this resource when you want to suppress input collection, and remove it when you're done.
```

src/plugin.rs
```diff
@@ -158,37 +113,25 @@ impl<A: Actionlike, UserState: Resource + Eq + Debug + Clone + Hash> Plugin
...
-            let input_manager_systems = input_manager_systems.with_run_criteria(
-                move |current_state: Res<State<UserState>>| {
-                    if *current_state.current() == desired_state_variant {
-                        ShouldRun::Yes
-                    } else {
-                        ShouldRun::No
-                    }
-                },
-            );
+/// A resource which disables all input for the specified `A` if present in world
```
alice-i-cecile, line 124:
```suggestion
/// A resource which disables all input for the specified [`Actionlike`] type `A` if present in world
```

tests/integration.rs
```diff
@@ -79,25 +79,16 @@ fn action_state_change_detection() {
 }
 
 #[test]
-fn run_in_state() {
+fn input_disables() {
```
alice-i-cecile, line 82:
```suggestion
fn disable_input() {
```

PR #107, opened 2022-04-07
Systems refactor

Shatur:


src/plugin.rs
```diff
@@ -69,53 +69,50 @@ impl<A: Actionlike> Plugin for InputManagerPlugin<A> {
...
-                            .label(InputManagerSystem::ReleaseOnDisable)
-                            .after(InputManagerSystem::Update),
-                    );
+                app.add_system_to_stage(
+                    CoreStage::PreUpdate,
+                    tick_action_state::<A>
+                        .label(InputManagerSystem::Tick)
+                        .before(InputManagerSystem::Update),
+                );
+                app.add_system_to_stage(
```
alice-i-cecile, line 80: Can you chain these instead?

Shatur: Oh, right,forgot about it :)

PR #109, opened 2022-04-11
Component resource Option

huhlig:
New attempt using components directly as a resource.


alice-i-cecile:
Looks good! Can you add a quick note to `RELEASES.md`? Then I think this is ready-to-go :)

RELEASES.md
```diff
@@ -4,6 +4,7 @@
 
 ### Enhancements
 
+- users are now able to use the `ActionState` Component as a resource externally to an entity
```
alice-i-cecile, line 7:
```suggestion
- if desired, users are now able to use the `ActionState` and `InputMap` structs as standalone resources
```

src/systems.rs
```diff
@@ -47,6 +55,8 @@ pub fn update_action_state<A: Actionlike>(
     maybe_keyboard_input_stream: Option<Res<Input<KeyCode>>>,
     maybe_mouse_input_stream: Option<Res<Input<MouseButton>>>,
     clash_strategy: Res<ClashStrategy>,
+    mut res_as: Option<ResMut<ActionState<A>>>,
+    mut res_im: Option<ResMut<InputMap<A>>>,
```
Shatur, line 59: Why are you using `res_` prefix? It's obvious that this is a resource. I would use `input_map` and `action_state`.

examples/binding_menu.rs
```diff
@@ -109,7 +110,7 @@ fn binding_window_system(
     mut input_events: InputEvents,
     active_binding: Option<ResMut<ActiveBinding>>,
     mut control_settings: ResMut<ControlSettings>,
-    ui_actions: Query<&ActionState<UiAction>>,
+    ui_actions: Res<ActionState<UiAction>>,
```
Shatur, line 113:
```suggestion
    ui_action_state: Res<ActionState<UiAction>>,
```

Shatur (changes requested):

src/systems.rs
huhlig: This was specifically to disambiguate the query from the resource. I had them as action_state_resource but it seemed too long. I can set them back to that if desired.

alice-i-cecile: Yeah, that's my preference here.

src/systems.rs
```diff
@@ -60,6 +70,20 @@ pub fn update_action_state<A: Actionlike>(
 
     let mouse = maybe_mouse_input_stream.as_deref();
 
+    if let (Some(input_map), Some(action_state)) =
+        (&mut input_map_resource, &mut action_state_resource)
```
huhlig, line 74: This was the kind of formatting I wanted to avoid but oh well.

Shatur: Sorry, but I meant don't use any suffix / prefix at all. In just makes variable longer for no reason.
For example, here In your code `Some(input_map), Some(action_state)` still resources, you just unwrapped the option.

I would rename `input_map_resource` into just `input_map`. In Rust it's okay to shadow varaible names with a different type to avoid having `something_option` or `something_str`.

src/systems.rs
```diff
@@ -24,11 +24,19 @@ pub fn tick_action_state<A: Actionlike>(
     mut query: Query<&mut ActionState<A>>,
     time: Res<Time>,
     disable_input: Option<Res<DisableInput<A>>>,
+    resource: Option<ResMut<ActionState<A>>>,
```
Shatur, line 27:
```suggestion
    action_state: Option<ResMut<ActionState<A>>>,
```

Shatur (changes requested):
Sorry, I probably should have explain myself more. Here is my suggestions.

PR #123, opened 2022-04-17
Revert rename

alice-i-cecile:


johanhelsing:
What was the reason you reverted the rename? (just curious)

alice-i-cecile:
I couldn't republish it correctly on crates.io :( Lesson learned: don't use hyphens in your crate names ever!

PR #129, opened 2022-04-28
Fix broken duration pressed functionality

### LINKED ISSUE #127: Press duration example not working - Previous Duration and Current Duration methods always return zero
## Version

0.3.0
Bevy 0.7.0

## Operating system & version
Windows 10

## What you did

I tried running the _press_duration_ example.

From my understating, the player object should "leap" when the left/right key is released, with the leap distance dependent on the length of the keypress.

## What actually happened

The player object didn't move at all.

## Additional information

I've narrowed down the problem to the _previous_duration_ and the _current_duration_ functions always returning zero. 

A minimal example of the issue (edited _minimal_ example)
```
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_plugin(InputManagerPlugin::<Action>::default())
        .add_startup_system(spawn_player)
        .add_system(jump)
        .run();
}

#[derive(Actionlike, PartialEq, Eq, Clone, Copy, Hash, Debug)]
enum Action {
    Run,
    Jump,
}

#[derive(Component)]
struct Player;

fn spawn_player(mut commands: Commands) {
    commands
        .spawn()
        .insert(Player)
        .insert_bundle(InputManagerBundle::<Action> {
            action_state: ActionState::default(),
            input_map: InputMap::new([(Action::Jump, KeyCode::Space)]),
        });
}
fn jump(query: Query<&ActionState<Action>, With<Player>>) {
    let action_state = query.single();
    if action_state.just_released(Action::Jump) {
        info!("End jump time: {:?}", action_state.previous_duration(Action::Jump));
    }
    if action_state.pressed(Action::Jump) {
        info!("Current jump time: {:?}", action_state.current_duration(Action::Jump));
    }
    if action_state.just_pressed(Action::Jump) {
        info!("Started jump!");
    }
}
```

alice-i-cecile:
Fixes #127.

Status: 
- [x] reproduce the bug
- [x] write an integration test
- [x] figure out why the integration test is failing
- [x] fix the bug

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,15 +35,13 @@ bevy_ui = {version = "0.7", default-features = false, optional = true}
 bevy_window = {version = "0.7", default-features = false}
 
 petitset = {version = "0.2.1", features = ["serde_compat"]}
-derive_more = "0.99"
-thiserror = "1.0"
+derive_more = {version = "0.99", default-features = false, features = ["display", "error"]}
 itertools = "0.10"
 serde = {version = "1.0", features = ["derive"]}
 
 [dev-dependencies]
 bevy = {version = "0.7", default-features = false, features = ["bevy_sprite", "bevy_text", "bevy_ui", "bevy_render", "bevy_core_pipeline", "x11"]}
 bevy_egui = {version="0.13", default-features = false}
-derive_more = "0.99"
 
 [lib]
 name = "leafwing_input_manager"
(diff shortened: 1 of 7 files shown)


alice-i-cecile:
Thoughts: 

1. Add debug invocations to `tick` and `flip`.
2. Try to figure out a path for exactly when the duration is getting changed; may be reset to zero.

PR #135, opened 2022-06-02
Store ToggleActions data in an enum

alice-i-cecile:
This is both prettier and clearer, but required some `PhantomData` shenanigans to make work.


alice-i-cecile:
The offending variant is hidden from docs like so: 

image

Shatur:
I think it looks better with a boolean variable as it was :)

alice-i-cecile:
Yeah I think you're right 😂

PR #137, opened 2022-06-02
Improve plugin-level docs

### LINKED ISSUE #6: Demonstrate how and why to use multiple action enums
Show an example with the player and camera each having a separate ActionState and InputMap.

### LINKED ISSUE #124: Erratic behavior without proper system ordering
## What problem did the developer encounter?

Without a proper system ordering, the behavior of `just_pressed()` is completely erratic, missing about 80% of keystrokes. This is because my own input handling system also runs in `CoreStage::PreUpdate`, so that I can handle inputs before the rest of the app's logic, and my system was not ordered with `leafwing_input_manager`'s own system(s).

## What was the fix?

```diff
  SystemSet::on_update(AppState::InGame)
      .with_system(inputs_system)
+     .after(InputManagerSystem::Update)
```

## How could this be better documented?

[Getting started](https://docs.rs/leafwing-input-manager/0.2.0/leafwing_input_manager/index.html#getting-started) does not mention system ordering, nor which stage this plugin runs at by default, nor the fact that there are 2 separate systems, one for resetting the input state, and one for updating it.

[`examles/minimal.rs`](https://github.com/Leafwing-Studios/leafwing_input_manager/blob/main/examples/minimal.rs) equally has no mention of system stage and ordering. In fact the only example using `.after()` is the ARPG one, and for a different unrelated reason as I can tell.

I appreciate my use case may be seen as already "advanced" as I moved a system to a different stage from the default one. However, because inputs generally need to be processed before other game/app logic, I would expect this to be a common pattern we should recommend anyway when using Bevy, and so it would be wise I think to discuss system ordering in the docs of an input plugin.

alice-i-cecile:
Fixes #124.

Fixes #6.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -10,6 +10,8 @@
 ### Docs
 
 - properly documented the `ToggleActions` functionality, for dynamically enabling and disabling actions
+- document the need to add system ordering when you have other functionality running during `CoreStage::PreUpdate`
+- hint to users that they may want to use multiple `Actionlike` enums
 
 ## Version 0.3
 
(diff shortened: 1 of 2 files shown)


djeedai:
Thanks!

PR #143, opened 2022-06-08
update crate name in README

ChristopherBiscardi:
heya! I was using leafwing today and got a warning when I copy/pasted the name from the readme instructions because the crate name uses dashes and not underscores.

Not sure if you want it to show underscores intentionally but figured I'd PR a change anyway.

thanks for the crate, it's very useful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ This ensures the examples are in-sync with the latest release.
 
 ### Getting started
 
-1. Add `leafwing_input_manager` to your `Cargo.toml`.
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
Ah right :( I tried and failed to update the name, but Cargo doesn't like name migrations. Lesson learned kids: never ever ever use hyphens in your crate names.

alice-i-cecile:
Thanks for the fix :)

PR #148, opened 2022-07-04
s/LCtrl/LControl

mwcz:
`LControl` [seems to be](https://github.com/bevyengine/bevy/blob/b1b3bd533bd76d25dfe0371c4bae0f5d31ebcd6c/crates/bevy_input/src/keyboard.rs#L301-L302) the name of the left Control variant.


alice-i-cecile (approved):
Thanks!

PR #150, opened 2022-07-05
Add Bevy Gilrs to Dev Dependency Features

zicklag:
This way you can run the examples with gamepad support. It was kind of confusing to me as somebody trying to run the examples that none of the gamepad inputs were working.

A lot of people probably wouldn't know to do this themselves.


alice-i-cecile:
Thanks!

Issue #50, opened 2022-01-27
Handle 2-dimensional analogue input


alice-i-cecile:
There are three main use patterns for this:

1. Partition the analog space into digital buttons
   - Handle with a mapping API that sets thresholds/boundaries for the values
2. Map analog values into actions, but still use the raw analog value for something (like smoothly changing walk/run on an analog stick)
   - Actually a special case of 1: You map onto an action, but then get back the analog value that generated that action
      - This can be done with reverse mapping, or with forward propagation
   - Destructuring this value ain't gonna happen
   - But we can return out the values from the virtual button if we store it
   - We actually don't need to return the value itself, just the user input that triggered it (since then the caller can look up the correct value themselves)
   - Requires #76
3. Mapping analog values to an action simply does not make sense (flight stick tilt)
   - Not our problem- can add helper features to `bevy_input` if necessary
      - In particular, "get a Vec2 from a particular axis" is an extremely common operation that is unnecessarily painful 

Eventually this should support:
- all gamepad axes
- touch emulated joysticks
- mouse emulated joysticks

Shatur:
Looking forward to it. Are you planning it for 0.2?

alice-i-cecile:
Probably not; it's a bit complex and I'd rather not block other important features and releases. We'll see though!

shanesveller:
Do you also consider `MouseWheel` events in scope here?

alice-i-cecile:
Hmm. That's a good question. Maybe? I suppose they could be used in similar ways.

Mostly though I think mouse wheel events act like a virtual button that can be held down instead though.

alice-i-cecile:
Completed.

PR #151, opened 2022-07-06
Add Support For Gamepad Axes

zicklag:
This should be functionally ready, but I haven't necessarily done my due diligence with the doc strings or self-review yet. I also need to add a new example demonstrating the use of the controller axes. I'll get to that tomorrow probably.

The PR ended up somewhat large, but the changes are pretty straight-forward.

- Now every event kind has a "value". This will be between -1.0 and 1.0.
- If you wanted to implement simple character bindings with an analog stick you could do this:
```rust
let input_map = InputMap::<PlayerAction>::new(vec![
    (
        PlayerAction::Right,
        GamepadAxisThreshold {
            axis: GamepadAxisType::LeftStickX,
            comparison: GamepadAxisComparison::Greater,
            threshold: 0.1,
        },
    ),
    (
        PlayerAction::Left,
        GamepadAxisThreshold {
            axis: GamepadAxisType::LeftStickX,
            comparison: GamepadAxisComparison::Less,
            threshold: -0.1,
        },
    ),
    (
        PlayerAction::Up,
        GamepadAxisThreshold {
            axis: GamepadAxisType::LeftStickY,
            comparison: GamepadAxisComparison::Greater,
            threshold: 0.1,
        },
    ),
    (
        PlayerAction::Down,
        GamepadAxisThreshold {
            axis: GamepadAxisType::LeftStickX,
            comparison: GamepadAxisComparison::Less,
            threshold: -0.1,
        },
    ),
]);
```
- In order to get the `f32` threshold to fit inside of the `Eq + Hash` `UserInput` struct, I made use of the `decorum` crate, which has suitable `Eq` and `Hash` implementations for floats. Technically we can get rid of the dependency if you want to copy the [`to_canonical_bits`](https://docs.rs/decorum/latest/src/decorum/canonical.rs.html#13-21) function from the `decorum` crate.

We really just need a way to hash and compare floats and that would give us what we need.

### DIFF
--- a/examples/axis_inputs.rs
+++ b/examples/axis_inputs.rs
@@ -77,7 +77,7 @@ fn move_player(query: Query<&ActionState<Action>, With<Player>>) {
     if action_state.pressed(Action::Move) {
         let axis_pair = action_state.action_axis_pair(Action::Move).unwrap();
         println!("Move:");
-        println!("   distance: {}", axis_pair.magnitude());
+        println!("   distance: {}", axis_pair.length());
         println!("          x: {}", axis_pair.x());
         println!("          y: {}", axis_pair.y());
     }
--- a/examples/axis_inputs.rs
+++ b/examples/axis_inputs.rs
@@ -14,11 +14,11 @@ fn main() {
         .run();
 }
 
-// This is the list of "things in the game I want to be able to do based on input"
 #[derive(Actionlike, PartialEq, Eq, Clone, Copy, Hash, Debug)]
 enum Action {
     Move,
     Throttle,
+    Rudder,
 }
 
 #[derive(Component)]
@@ -32,19 +32,35 @@ fn spawn_player(mut commands: Commands) {
             // Stores "which actions are currently activated"
             action_state: ActionState::default(),
             // Describes how to convert from player inputs into those actions
-            input_map: InputMap::new([(
-                // We want to trigger our move action when the left stick is moved more than 10% in
-                // any direction.
-                DualGamepadAxisThreshold {
-                    x_axis: GamepadAxisType::LeftStickX,
-                    y_axis: GamepadAxisType::LeftStickY,
-                    comparison: GamepadAxisComparison::Greater,
-                    threshold: 0.1,
-                },
-                Action::Move,
-            )])
+            input_map: InputMap::new([
+                // Configure the left stick as a dual-axis
+                (
+                    DualGamepadAxis {
+                        x_axis: GamepadAxisType::LeftStickX,
+                        y_axis: GamepadAxisType::LeftStickY,
+                        // We want to trigger our move action when the left stick is moved more than 10%
+                        // in any direction.
+                        //
+                        // Note: The Bevy `GamepadSettings` such as the axis deadzone with still apply.
+                        // If bevy filters out a gamepad event,
+                        // it will be filtered out of leafwing, too.
+                        deadzone: 0.1,
+                    },
+                    Action::Move,
+                ),
+            ])
             // Let's also add a gamepad button binding to the right trigger
             .insert(GamepadButtonType::RightTrigger2, Action::Throttle)
+            // And we'll use the x axis on the right stick as a rudder control
+            .insert(
+                SingleGamepadAxis {
+                    axis: GamepadAxisType::RightStickX,
+                    // This will trigger if the axis is moved 10% or more in either direction.
+                    negative_low: -0.1,
+                    positive_low: 0.1,
+                },
+                Action::Rudder,
+            )
             // Listen for events on the first gamepad
             .set_gamepad(Gamepad(0))
             .build(),
@@ -56,7 +72,7 @@ fn move_player(query: Query<&ActionState<Action>, With<Player>>) {
     let action_state = query.single();
     // Each action has a button-like state of its own that you can check
     if action_state.pressed(Action::Move) {
-        let axis_pair = action_state.action_axis_pair(Action::Move);
+        let axis_pair = action_state.action_axis_pair(Action::Move).unwrap();
         println!("Move:");
         println!("   distance: {}", axis_pair.magnitude());
         println!("          x: {}", axis_pair.x());
@@ -73,5 +89,10 @@ fn move_player(query: Query<&ActionState<Action>, With<Player>>) {
         // this will just return 0.0 when not pressed and 1.0 when pressed.
         let value = action_state.action_value(Action::Throttle);
         println!("Throttle: {}", value);
+    }
+
+    if action_state.pressed(Action::Rudder) {
+        let value = action_state.action_value(Action::Rudder);
+        println!("Rudder: {}", value);
     }
 }
--- a/examples/axis_inputs.rs
+++ b/examples/axis_inputs.rs
@@ -41,9 +41,8 @@ fn spawn_player(mut commands: Commands) {
                         // We want to trigger our move action when the left stick is moved more than 10%
                         // in any direction.
                         //
-                        // Note: The Bevy `GamepadSettings` such as the axis deadzone with still apply.
-                        // If bevy filters out a gamepad event,
-                        // it will be filtered out of leafwing, too.
+                        // Note: The Bevy `GamepadSettings` may cause input to be filtered out before
+                        // reaching leafwing.
                         x_positive_low: 0.1,
                         x_negative_low: -0.1,
                         y_positive_low: 0.1,
@@ -52,9 +51,9 @@ fn spawn_player(mut commands: Commands) {
                     Action::Move,
                 ),
             ])
-            // Let's also add a gamepad button binding to the right trigger
+            // Let's bind the right gamepad trigger to the throttle action
             .insert(GamepadButtonType::RightTrigger2, Action::Throttle)
-            // And we'll use the x axis on the right stick as a rudder control
+            // And we'll use the right stick's x axis as a rudder control
             .insert(
                 SingleGamepadAxis {
                     axis: GamepadAxisType::RightStickX,
@@ -83,13 +82,12 @@ fn move_player(query: Query<&ActionState<Action>, With<Player>>) {
     }
 
     if action_state.pressed(Action::Throttle) {
-        // Note that some gamepad buttons are also tied to axes,
-        // so even though we used a GamepadbuttonType::RightTrigger2 binding to trigger the
-        // throttle action,
-        // we can get a variable value here if you have a variable right trigger on your gamepad.
+        // Note that some gamepad buttons are also tied to axes, so even though we used a
+        // GamepadbuttonType::RightTrigger2 binding to trigger the throttle action, we can get a
+        // variable value here if you have a variable right trigger on your gamepad.
         //
-        // If you don't have a variable trigger,
-        // this will just return 0.0 when not pressed and 1.0 when pressed.
+        // If you don't have a variable trigger, this will just return 0.0 when not pressed and 1.0
+        // when pressed.
         let value = action_state.action_value(Action::Throttle);
         println!("Throttle: {}", value);
     }
--- a/examples/axis_inputs.rs
+++ b/examples/axis_inputs.rs
@@ -44,7 +44,10 @@ fn spawn_player(mut commands: Commands) {
                         // Note: The Bevy `GamepadSettings` such as the axis deadzone with still apply.
                         // If bevy filters out a gamepad event,
                         // it will be filtered out of leafwing, too.
-                        deadzone: 0.1,
+                        x_positive_low: 0.1,
+                        x_negative_low: -0.1,
+                        y_positive_low: 0.1,
+                        y_negative_low: -0.1,
                     },
                     Action::Move,
                 ),
--- a/examples/virtual_dpad.rs
+++ b/examples/virtual_dpad.rs
@@ -31,7 +31,7 @@ fn spawn_player(mut commands: Commands) {
             action_state: ActionState::default(),
             // Describes how to convert from player inputs into those actions
             input_map: InputMap::new([(
-                UserInput::VirtualDPad {
+                VirtualDPad {
                     up: KeyCode::W.into(),
                     down: KeyCode::S.into(),
                     left: KeyCode::A.into(),
@@ -50,7 +50,7 @@ fn move_player(query: Query<&ActionState<Action>, With<Player>>) {
     if action_state.pressed(Action::Move) {
         let axis_pair = action_state.action_axis_pair(Action::Move).unwrap();
         println!("Move:");
-        println!("   distance: {}", axis_pair.magnitude());
+        println!("   distance: {}", axis_pair.length());
         println!("          x: {}", axis_pair.x());
         println!("          y: {}", axis_pair.y());
     }
--- a/examples/virtual_dpad.rs
+++ b/examples/virtual_dpad.rs
@@ -14,7 +14,6 @@ fn main() {
         .run();
 }
 
-// This is the list of "things in the game I want to be able to do based on input"
 #[derive(Actionlike, PartialEq, Eq, Clone, Copy, Hash, Debug)]
 enum Action {
     Move,
@@ -49,7 +48,7 @@ fn move_player(query: Query<&ActionState<Action>, With<Player>>) {
     let action_state = query.single();
     // Each action has a button-like state of its own that you can check
     if action_state.pressed(Action::Move) {
-        let axis_pair = action_state.action_axis_pair(Action::Move);
+        let axis_pair = action_state.action_axis_pair(Action::Move).unwrap();
         println!("Move:");
         println!("   distance: {}", axis_pair.magnitude());
         println!("          x: {}", axis_pair.x());
--- a/examples/virtual_dpad.rs
+++ b/examples/virtual_dpad.rs
@@ -29,7 +29,7 @@ fn spawn_player(mut commands: Commands) {
         .insert_bundle(InputManagerBundle::<Action> {
             // Stores "which actions are currently activated"
             action_state: ActionState::default(),
-            // Describes how to convert from player inputs into those actions
+            // Map some arbitrary keys into a virtual direction pad that triggers our move action
             input_map: InputMap::new([(
                 VirtualDPad {
                     up: KeyCode::W.into(),
@@ -46,8 +46,10 @@ fn spawn_player(mut commands: Commands) {
 // Query for the `ActionState` component in your game logic systems!
 fn move_player(query: Query<&ActionState<Action>, With<Player>>) {
     let action_state = query.single();
-    // Each action has a button-like state of its own that you can check
+    // If any button in a virtual direction pad is pressed, then the action state is "pressed"
     if action_state.pressed(Action::Move) {
+        // Virtual direction pads are one of the types which return an AxisPair. The values will be
+        // represented as `-1.0`, `0.0`, or `1.0` depending on the combination of buttons pressed.
         let axis_pair = action_state.action_axis_pair(Action::Move).unwrap();
         println!("Move:");
         println!("   distance: {}", axis_pair.length());
--- a/examples/virtual_dpad.rs
+++ b/examples/virtual_dpad.rs
@@ -18,7 +18,6 @@ fn main() {
 #[derive(Actionlike, PartialEq, Eq, Clone, Copy, Hash, Debug)]
 enum Action {
     Move,
-    Throttle,
 }
 
 #[derive(Component)]
@@ -33,20 +32,14 @@ fn spawn_player(mut commands: Commands) {
             action_state: ActionState::default(),
             // Describes how to convert from player inputs into those actions
             input_map: InputMap::new([(
-                // We want to trigger our move action when the left stick is moved more than 10% in
-                // any direction.
-                DualGamepadAxisThreshold {
-                    x_axis: GamepadAxisType::LeftStickX,
-                    y_axis: GamepadAxisType::LeftStickY,
-                    comparison: GamepadAxisComparison::Greater,
-                    threshold: 0.1,
+                UserInput::VirtualDPad {
+                    up: KeyCode::W.into(),
+                    down: KeyCode::S.into(),
+                    left: KeyCode::A.into(),
+                    right: KeyCode::D.into(),
                 },
                 Action::Move,
             )])
-            // Let's also add a gamepad button binding to the right trigger
-            .insert(GamepadButtonType::RightTrigger2, Action::Throttle)
-            // Listen for events on the first gamepad
-            .set_gamepad(Gamepad(0))
             .build(),
         });
 }
@@ -61,17 +54,5 @@ fn move_player(query: Query<&ActionState<Action>, With<Player>>) {
         println!("   distance: {}", axis_pair.magnitude());
         println!("          x: {}", axis_pair.x());
         println!("          y: {}", axis_pair.y());
-    }
-
-    if action_state.pressed(Action::Throttle) {
-        // Note that some gamepad buttons are also tied to axes,
-        // so even though we used a GamepadbuttonType::RightTrigger2 binding to trigger the
-        // throttle action,
-        // we can get a variable value here if you have a variable right trigger on your gamepad.
-        //
-        // If you don't have a variable trigger,
-        // this will just return 0.0 when not pressed and 1.0 when pressed.
-        let value = action_state.action_value(Action::Throttle);
-        println!("Throttle: {}", value);
     }
 }
--- a/src/action_state.rs
+++ b/src/action_state.rs
@@ -195,19 +195,20 @@ impl<A: Actionlike> ActionState<A> {
     ///
     /// Different kinds of bindings have different ways of calculating the value:
     ///
-    /// - Binary buttons will have a value of `0.0` when the button is not pressed,
-    /// and a value of `1.0` when the button is pressed.
+    /// - Binary buttons will have a value of `0.0` when the button is not pressed, and a value of
+    /// `1.0` when the button is pressed.
     /// - Some axes, such as an analog stick, will have a value in the range `-1.0..=1.0`.
     /// - Some axes, such as a variable trigger, will have a value in the range `0.0..=1.0`.
-    /// - Some buttons will also return a value in the range `0.0..=1.0`,
-    /// such as variable triggers, which may be tracked as buttons or axes.
-    /// - Dual axis inputs will return the magnitude of it's [`AxisPair`] and will be in the range
+    /// - Some buttons will also return a value in the range `0.0..=1.0`, such as analog gamepad
+    /// triggers which may be tracked as buttons or axes. Examples of these include the Xbox LT/RT
+    /// triggers and the Playstation L2/R2 triggers. See also the `axis_inputs` example in the
+    /// repository.
+    /// - Dual axis inputs will return the magnitude of its [`AxisPair`] and will be in the range
     /// `0.0..=1.0`.
-    /// - Chord inputs will return the value of it's first input.
+    /// - Chord inputs will return the value of its first input.
     ///
-    /// If multiple inputs trigger the same game action at the same time,
-    /// the value of each triggering input will be added together and clamped to the range
-    /// `-1.0..=1.0`
+    /// If multiple inputs trigger the same game action at the same time, the value of each
+    /// triggering input will be added together and clamped to the range `-1.0..=1.0`
     pub fn action_value(&self, action: A) -> f32 {
         self.action_data(action).value
     }
--- a/src/action_state.rs
+++ b/src/action_state.rs
@@ -214,14 +214,14 @@ impl<A: Actionlike> ActionState<A> {
 
     /// Get the [`AxisPair`] from the binding that triggered the corresponding `action`.
     ///
-    /// Only certain events such as [`UserInput::VirtualDPad`] and
-    /// [`InputKind::DualGamepadAxis`][crate::user_input::InputKind::DualGamepadAxis] provide an
-    /// [`AxisPair`], and this will return [`None`] for other events.
+    /// Only certain events such as [`VirtualDPad`][crate::user_input::VirtualDPad] and
+    /// [`DualGamepadAxis`][crate::user_input::DualGamepadAxis] provide an [`AxisPair`], and this
+    /// will return [`None`] for other events.
     ///
     /// Chord inputs will return the [`AxisPair`] of it's first input.
     ///
-    /// If multiple inputs with an axis pair trigger the same game action at the same time,
-    /// the value of each axis pair will be added together and be clamped to a maximum magnitude of
+    /// If multiple inputs with an axis pair trigger the same game action at the same time, the
+    /// value of each axis pair will be added together and be clamped to a maximum magnitude of
     /// `1.0`.
     pub fn action_axis_pair(&self, action: A) -> Option<AxisPair> {
         self.action_data(action).axis_pair
--- a/src/action_state.rs
+++ b/src/action_state.rs
@@ -16,6 +16,11 @@ use std::marker::PhantomData;
 pub struct ActionData {
     /// Is the action pressed or released?
     pub state: ButtonState,
+    /// The analog value of this button/axis, if applicable
+    ///
+    /// For binary inputs such as buttons, this will always be either 0.0 or 1.0. For analog axes
+    /// this will be in the range of `-1.0..=1.0`.
+    pub value: f32,
     /// What inputs were responsible for causing this action to be pressed?
     pub reasons_pressed: Vec<UserInput>,
     /// When was the button pressed / released, and how long has it been held for?
@@ -96,6 +101,7 @@ impl<A: Actionlike> ActionState<A> {
                 ButtonState::Released => self.release(action),
             }
 
+            self.action_data[i].value = action_data[i].value;
             self.action_data[i].reasons_pressed = action_data[i].reasons_pressed.clone();
         }
     }
--- a/src/action_state.rs
+++ b/src/action_state.rs
@@ -204,7 +204,7 @@ impl<A: Actionlike> ActionState<A> {
     /// - Dual axis inputs will return the magnitude of it's [`AxisPair`] and will be in the range
     /// `0.0..=1.0`.
     /// - Chord inputs will return the value of it's first input.
-    /// 
+    ///
     /// If multiple inputs trigger the same game action at the same time,
     /// the value of each triggering input will be added together and clamped to the range
     /// `-1.0..=1.0`
@@ -217,9 +217,9 @@ impl<A: Actionlike> ActionState<A> {
     /// Only certain events such as [`UserInput::VirtualDPad`] and
     /// [`InputKind::DualGamepadAxis`][crate::user_input::InputKind::DualGamepadAxis] provide an
     /// [`AxisPair`], and this will return [`None`] for other events.
-    /// 
+    ///
     /// Chord inputs will return the [`AxisPair`] of it's first input.
-    /// 
+    ///
     /// If multiple inputs with an axis pair trigger the same game action at the same time,
     /// the value of each axis pair will be added together and be clamped to a maximum magnitude of
     /// `1.0`.
--- a/src/action_state.rs
+++ b/src/action_state.rs
@@ -187,6 +187,23 @@ impl<A: Actionlike> ActionState<A> {
         self.action_data[action.index()].clone()
     }
 
+    /// Get the value associated with the corresponding `action`
+    /// 
+    /// For actions triggered by a variable axis, such as an analog stick,
+    /// this will be in the range of `-1.0..=1.0`.
+    /// 
+    /// For buttons this will be `0.0` when the button is not pressed,
+    /// and `1.0` when the button is pressed.
+    /// 
+    /// Some axes such as trigger buttons will return a value in the range of `-1.0..=1.0`.
+    /// 
+    /// Note that some buttons, such as [`GamepadButtonType::RightTrigger2`] are variable.
+    /// Variable buttons will return how "pressed" they are in this case as a value in the range
+    /// `0.0..=1.0`.
+    pub fn action_value(&self, action: A) -> f32 {
+        self.action_data(action).value
+    }
+
     /// Manually sets the [`ActionData`] of the corresponding `action`
     ///
     /// You should almost always use more direct methods, as they are simpler and less error-prone.
--- a/src/action_state.rs
+++ b/src/action_state.rs
@@ -1,8 +1,8 @@
 //! This module contains [`ActionState`] and its supporting methods and impls.
 
-use crate::buttonlike::ButtonState;
 use crate::user_input::UserInput;
 use crate::Actionlike;
+use crate::{axislike::AxisPair, buttonlike::ButtonState};
 
 use bevy_ecs::{component::Component, entity::Entity};
 use bevy_utils::{Duration, Instant};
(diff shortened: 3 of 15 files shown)


alice-i-cecile:
> We really just need a way to hash and compare floats and that would give us what we need.

Take a look at https://github.com/bevyengine/bevy/pull/5208 and the underlying `FloatOrd` type. Perhaps just import it?

zicklag:
That looks like it would work for the ordering, but it won't work for the hashing unfortunately. I can't find anything in Bevy or the standard library that will hash a float. :/

alice-i-cecile:
It impls `Hash` too :) https://github.com/bevyengine/bevy/blob/8765658acf5840469692edfaa102378e3c8dedc4/crates/bevy_utils/src/float_ord.rs#L32

zicklag:
Oh! Good find. Now I've got to make a PR for the docs to mention that. :)

That should work fine.

zicklag:
@alice-i-cecile do you think we should change the name of `just_pressed` to `just_activated` now that actions aren't always related to literal "presses"? Or maybe we make `just_pressed` a deprecated alias of `just_activated`.

A quick search yields 279 ocurrences of the term "pressed" throughout the codebase, so there'd probably be a lot of re-naming if we wanted to make it all consistent.

Either way, that should be a separate PR, but was curious if that's something you think would make sense.

alice-i-cecile:
> do you think we should change the name of just_pressed to just_activated now that actions aren't always related to literal "presses"

IMO we should aim for consistency with `bevy_input`, which uses `just_pressed` everywhere. I tend to think of this sort of thing as using "virtual buttons".

src/user_input.rs
```diff
@@ -240,12 +249,12 @@ impl Iterator for InputModeIter {
...
 
-impl From<InputButton> for InputMode {
-    fn from(button: InputButton) -> Self {
+impl From<InputKind> for InputMode {
+    fn from(button: InputKind) -> Self {
         match button {
-            InputButton::Gamepad(_) => InputMode::Gamepad,
-            InputButton::Keyboard(_) => InputMode::Keyboard,
-            InputButton::Mouse(_) => InputMode::Mouse,
+            InputKind::GamepadButton(_) | InputKind::GamepadAxis(_) => InputMode::Gamepad,
```
zicklag, line 255: I'm not 100% what `InputMode` is meant to indicate. Should we have `GamepadButton` and `GamepadAxis` variants of `InputMode`, or just `Gamepad`?

src/user_input.rs
```diff
@@ -260,30 +269,78 @@ impl From<InputButton> for InputMode {
 /// Please contact the maintainers if you need support for another type!
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub enum InputButton {
+pub enum InputKind {
```
zicklag, line 272: I renamed this, but would you rather keep it `InputButton` in favor of the "virtual button" mentality?

zicklag (commented):
OK, I think this is ready for review now, and I left a couple questions.

CleanCut:
Hey, friends! I've got a couple questions:

- A couple axes (`GamepadAxisType::RightZ` and `GamepadAxisType::LeftZ`) typically present values in a range from `0.0..=1.0` (instead of `-1.0..=1.0` like most axes). Q: Is that handled here?
- I see that `GamepadAxisThreshold` converts axis input to button-like behavior (a better name may be something like `GamepadAxisButton` or something like that). That is great for folks who want to treat an axis as a button. Q: Where is the `impl UserInput` that allows treating an axis like an axis?

@zicklag I'm willing to pitch in and help work on this PR with you if you like. 😄

src/user_input.rs
CleanCut: I'd like to be able to map a gamepad axis to an action axis as well (not _button_). `InputKind` seems like a decent name if it were going to include both.

src/user_input.rs
CleanCut: As far as I can tell, `InputMode` is used when you want to take some input and simplify it into where it came from. If that is correct, then you're doing the right thing here, as both the gamepad buttons and axes are parts of a gamepad.

src/user_input.rs
CleanCut: But it _might_ make more sense to have separate `InputButton` and `InputAxis` enums, especially if we want to be able to add non-gamepad axes in the future (like a mouse-wheel axis, or a touch-drag axis, etc.).

zicklag: > InputKind seems like a decent name if it were going to include both buttons and axes in a single enum.

It seemed more intuitive to me, and I was thinking that maybe it would prevent people from being turned off from investigating the type if they were looking for the Axis inputs.

> But it might make more sense to have separate InputButton and InputAxis enums, especially if we want to be able to add non-gamepad axes in the future (like a mouse-wheel axis, or a touch-drag axis, etc.).

I _think_ that it's best to keep all that in one enum, because the rest of the library is kind of centered around it, and splitting it out would make things more difficult functionally. But I could see adding more variants to it as we expand the kinds of inputs that we accept.

zicklag:
Hey @CleanCut thanks for the review! As for you questions:

> A couple axes (GamepadAxisType::RightZ and GamepadAxisType::LeftZ) typically present values in a range from 0.0..=1.0 (instead of -1.0..=1.0 like most axes). Q: Is that handled here?

Those axes are handled as you would expect. If Bevy returns a value in the range `0.0..=1.0` for that axis, then that is what you will get when reading the value. I just updated the `analog_stick` example that I added in this PR so that it shows how to read the axis tilt amount.

> I see that GamepadAxisThreshold converts axis input to button-like behavior (a better name may be something like GamepadAxisButton or something like that). That is great for folks who want to treat an axis as a button. Q: Where is the impl UserInput that allows treating an axis like an axis?

Even if you use a `GamepadAxisThreshold` to define a binding, you can still use it as an axis. We just have to have some criteria for when to trigger the action. I did just realize there could be a use-case for _always_ triggering an action, and I can add that easily.  That would let you get the value of the axis at any time in the same way that you get the value in the `analog_stick` example.

---

Also, that just gave me an idea for how to handle dual axes and now I want to try that out. :) I'm going to do a quick update so I can show you guys what I'm thinking for that and get your thoughts on it.

> @zicklag I'm willing to pitch in and help work on this PR with you if you like. smile

Sure thing! I just want to get the next push of dual-axis handling up first and then we can go from there. If it doesn't look right and we want to reduce scope then, we can just rip it out and get single axes working first.

src/user_input.rs
alice-i-cecile: Precisely.

zicklag:
OK, I've got a quick draft of dual axis support! The best way to explain is with the example, which I'll post in full below to avoid broken links from new commits, etc.

<details>

```rust
use bevy::prelude::*;
use leafwing_input_manager::{prelude::*, user_input::DualGamepadAxisThreshold};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        // This plugin maps inputs to an input-type agnostic action-state
        // We need to provide it with an enum which stores the possible actions a player could take
        .add_plugin(InputManagerPlugin::<Action>::default())
        // Spawn an entity with Player, InputMap, and ActionState components
        .add_startup_system(spawn_player)
        // Read the ActionState in your systems using queries!
        .add_system(move_player)
        .run();
}

// This is the list of "things in the game I want to be able to do based on input"
#[derive(Actionlike, PartialEq, Eq, Clone, Copy, Hash, Debug)]
enum Action {
    Move,
    Throttle,
}

#[derive(Component)]
struct Player;

fn spawn_player(mut commands: Commands) {
    commands
        .spawn()
        .insert(Player)
        .insert_bundle(InputManagerBundle::<Action> {
            // Stores "which actions are currently activated"
            action_state: ActionState::default(),
            // Describes how to convert from player inputs into those actions
            input_map: InputMap::new([(
                // We want to trigger our move action when the left stick is moved more than 10% in
                // any direction.
                DualGamepadAxisThreshold {
                    x_axis: GamepadAxisType::LeftStickX,
                    y_axis: GamepadAxisType::LeftStickY,
                    comparison: GamepadAxisComparison::Greater,
                    threshold: 0.1,
                },
                Action::Move,
            )])
            // Let's also add a gamepad button binding to the right trigger
            .insert(GamepadButtonType::RightTrigger2, Action::Throttle)
            // Listen for events on the first gamepad
            .set_gamepad(Gamepad(0))
            .build(),
        });
}

// Query for the `ActionState` component in your game logic systems!
fn move_player(query: Query<&ActionState<Action>, With<Player>>) {
    let action_state = query.single();
    // Each action has a button-like state of its own that you can check
    if action_state.pressed(Action::Move) {
        let axis_pair = action_state.action_axis_pair(Action::Move);
        println!("Move:");
        println!("   distance: {}", axis_pair.magnitude());
        println!("          x: {}", axis_pair.x());
        println!("          y: {}", axis_pair.y());
    }

    if action_state.pressed(Action::Throttle) {
        // Note that some gamepad buttons are also tied to axes,
        // so even though we used a GamepadbuttonType::RightTrigger2 binding to trigger the
        // throttle action,
        // we can get a variable value here if you have a variable right trigger on your gamepad.
        //
        // If you don't have a variable trigger,
        // this will just return 0.0 when not pressed and 1.0 when pressed.
        let value = action_state.action_value(Action::Throttle);
        println!("Throttle: {}", value);
    }
}
```

</details>

The PR might be a little rough, I didn't comb through it again for better doc comments, consistency, etc. but I did attempt to put decent explanations on the new `ActionState::action_value()` and `ActionState::action_axis_pair()` methods.

I'd be great to hear your thoughts on this @CleanCut to see if this gets you what you're looking for.

zicklag:
I just added support for virtual DPads.

This PR is pretty bloated, but the changes are in separate commits so I can easily split them out if you want. I think the biggest point right now is to get review on the public API, and if that looks good, we can butcher this PR 'till it's reviewable/mergable.

Here's how to use the virtual DPad, which works with both keyboard keys like WASD or with gamepad buttons such as the DPad directions:

<details>

```rust
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        // This plugin maps inputs to an input-type agnostic action-state
        // We need to provide it with an enum which stores the possible actions a player could take
        .add_plugin(InputManagerPlugin::<Action>::default())
        // Spawn an entity with Player, InputMap, and ActionState components
        .add_startup_system(spawn_player)
        // Read the ActionState in your systems using queries!
        .add_system(move_player)
        .run();
}

// This is the list of "things in the game I want to be able to do based on input"
#[derive(Actionlike, PartialEq, Eq, Clone, Copy, Hash, Debug)]
enum Action {
    Move,
}

#[derive(Component)]
struct Player;

fn spawn_player(mut commands: Commands) {
    commands
        .spawn()
        .insert(Player)
        .insert_bundle(InputManagerBundle::<Action> {
            // Stores "which actions are currently activated"
            action_state: ActionState::default(),
            // Describes how to convert from player inputs into those actions
            input_map: InputMap::new([(
                UserInput::VirtualDPad {
                    up: KeyCode::W.into(),
                    down: KeyCode::S.into(),
                    left: KeyCode::A.into(),
                    right: KeyCode::D.into(),
                },
                Action::Move,
            )])
            // Listen for events on the first gamepad
            .set_gamepad(Gamepad(0))
            .build(),
        });
}

// Query for the `ActionState` component in your game logic systems!
fn move_player(query: Query<&ActionState<Action>, With<Player>>) {
    let action_state = query.single();
    // Each action has a button-like state of its own that you can check
    if action_state.pressed(Action::Move) {
        let axis_pair = action_state.action_axis_pair(Action::Move);
        println!("Move:");
        println!("   distance: {}", axis_pair.magnitude());
        println!("          x: {}", axis_pair.x());
        println!("          y: {}", axis_pair.y());
    }
}
```

</details>

examples/virtual_dpad.rs
```diff
@@ -0,0 +1,58 @@
...
+        // We need to provide it with an enum which stores the possible actions a player could take
+        .add_plugin(InputManagerPlugin::<Action>::default())
+        // Spawn an entity with Player, InputMap, and ActionState components
+        .add_startup_system(spawn_player)
+        // Read the ActionState in your systems using queries!
+        .add_system(move_player)
+        .run();
+}
+
+// This is the list of "things in the game I want to be able to do based on input"
```
alice-i-cecile, line 17:
```suggestion
```

src/input_map.rs
```diff
@@ -312,11 +313,15 @@ impl<A: Actionlike> InputMap<A> {
...
+                let value = input_streams.get_input_value(input);
+                let axis_pair = input_streams.get_input_axis_pair(input);
                 if input_streams.input_pressed(input) {
                     inputs.push(input.clone());
-                    action_data[action.index()]
-                        .reasons_pressed
-                        .push(input.clone());
+                    let action = &mut action_data[action.index()];
+                    action.reasons_pressed.push(input.clone());
+                    action.value += value;
```
alice-i-cecile, line 322: Interesting, so then you can handle multiple overlapping inputs simultaneously 🤔 Very nice.

src/user_input.rs
```diff
@@ -1,444 +1,827 @@
...
+};
+
+use bevy_math::Vec2;
+use bevy_utils::HashSet;
+use petitset::PetitSet;
+use serde::{Deserialize, Serialize};
+
+use crate::axislike::AxisPair;
+
+/// Some combination of user input, which may cross [`Input`] boundaries
```
alice-i-cecile, line 18:
```suggestion
/// Some combination of user input, which may cross [`Input`]-mode boundaries
```

src/user_input.rs
```diff
@@ -1,444 +1,827 @@
...
+pub enum UserInput {
+    /// A single button
+    Single(InputKind),
+    /// A combination of buttons, pressed simultaneously
+    ///
+    /// Up to 8 (!!) buttons can be chorded together at once.
+    /// Chords are considered to belong to all of the [InputMode]s of their constituent buttons.
+    Chord(PetitSet<InputKind, 8>),
+    /// A virtual DPad that you can get an [`AxisPair`] from
+    VirtualDPad {
```
alice-i-cecile, line 31: Oh that's a very nice design. This pattern comes up a lot!

src/user_input.rs
```diff
@@ -1,444 +1,827 @@
...
+    left: InputKind,
+    right: InputKind,
+}
+
+/// Different possible ways to compare an axis threshold to the axis value.
+///
+/// See [`GamepadAxisThreshold`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum GamepadAxisComparison {
+    /// Value must be less than threshold.
```
alice-i-cecile, line 454: Presumably these are both "or equal to"?

alice-i-cecile (changes requested):
First round of review from me. Looking very good, but there's a couple nits.

The other major thing to add is that we should add a section on this to the `RELEASES.md` file :D

src/action_state.rs
```diff
@@ -16,6 +16,14 @@ use std::marker::PhantomData;
...
     /// Is the action pressed or released?
     pub state: ButtonState,
+    /// The "value" of the binding that triggered the action.
+    ///
+    /// See [`ActionState::action_value()`] for more details.
+    pub value: f32,
+    /// The [`AxisPair`] of the binding that triggered the action.
+    ///
+    /// See [`ActionState::action_axis_pair()`] for more details.
+    pub axis_pair: AxisPair,
```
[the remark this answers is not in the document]
zicklag: Yeah, now that I think about it, that makes sense. Otherwise there's no way to tell whether the AxisPair value returned is actually meaningful or not.

src/user_input.rs
```diff
@@ -1,444 +1,827 @@
...
+    down: InputKind,
+    left: InputKind,
+    right: InputKind,
+}
+
+/// Different possible ways to compare an axis threshold to the axis value.
+///
+/// See [`GamepadAxisThreshold`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum GamepadAxisComparison {
```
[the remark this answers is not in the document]
zicklag: For `Equal` the use-case I could imagine was if you wanted an action to be triggered when the joystick was at `0.0`. It was easy to implement and not confusing to the user so I threw it in there, but I'm not sure if we really need it.

For `Always` the idea was if you needed to monitor an axis constantly, but you still didn't know which axis, because the user was allowed to configure the input mapping. I think that's a valid use-case. That way you can still query the axis value, similar to how you would monitor an axis in normal Bevy, but you still get the input abstraction where you don't have to know exactly which axis / virtual DPad / etc. triggered the input.

Otherwise you could emulate it by triggering less than `10000000`, or something like that, but that seems sloppy.

Still, I don't personally have a use-case for it yet, so maybe we leave it out unless somebody asks for it?

@CleanCut do you have any thoughts on this one?

zicklag (commented):
Thanks for the review @alice-i-cecile! I'll go address the points you made and do another look over it myself. I think I want to add a demonstration of the SingleGamepadAxis to the analog_stick example, and I think there were a couple doc comments to make more thorough.

src/user_input.rs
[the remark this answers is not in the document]
zicklag: I just ran into the need for an `AbsoluteGreater` variant, to trigger if the axis movement was greater than a certain amount in any direction.

Do you think that `AbsoluteGreater` and `AbsoluteLess` variants make sense?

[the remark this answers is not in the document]
CleanCut: > @CleanCut do you have any thoughts on this one?

I don't see a need for either `Equal` or `Always`.

> Do you think that AbsoluteGreater and AbsoluteLess variants make sense?

Given that `bevy_input` has a way to configure dead zones so that they're filtered out at that level AND the fact that `bevy_input` uses the term threshold to mean "the minimum difference between old and new values to apply the changes", I'm wondering if we should be managing deadzones at this level?  Maybe we don't need to? If we do, we should probably use the same terminology as `bevy_input`.

```
// the bevy_input way...
fn configure_gamepads(mut settings: ResMut<GamepadSettings>) {
    // add a larger default dead-zone to all axes (ignore small inputs, round to zero)
    settings.default_axis_settings.negative_low = -0.125;
    settings.default_axis_settings.positive_low = 0.125;

    // the axis value must change this much before it will register as a different value
    settings.default_axis_settings.threshold = 0.05;
}
```

CleanCut:
I'm not getting consistent axis values through the plugin using this PR.  When two axes are both near max, they stop registering the same as what I'm getting straight from `bevy_input`.

I use the _left_ stick X to turn and the _right_ stick Y to accelerate. My "flame" sprite is set each frame based on the Y axis value of the _right_ stick. I get missed input while holding my right stick all the way up and _then_ turning with the left stick nearly all the way (it only starts fritzing out when both axis are near max value).

Raw bevy_input (good):

https://user-images.githubusercontent.com/5838512/177843164-01ea6b80-5631-4b02-962d-157ea51b4d47.mp4


Leafwing with this PR (bad):


https://user-images.githubusercontent.com/5838512/177843182-48127f4c-e431-47ab-a083-be78a105bc69.mp4

src/user_input.rs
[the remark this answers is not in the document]
alice-i-cecile: I like the absolute variants; feel free to add those.

Seconding @CleanCut on the "should we just be using bevy_inputs" settings though :)

CleanCut:
I figured out what the difference was. With `bevy_input` and separate axes, I would get `(-1.0, 1.0)` if I were turning full left at full throttle, while with leafwing the two separate axes were combined into a single axis and normalized to unit length, so I would get `(-0.70710677, 0.70710677)` with leafwing.  The difference caused my calculations to land right near the border for triggering the full flame, so the flickering was due to the slight variations in frame rate causing the value to wander back and forth over the border (since the value I was monitoring had been multiplied by `time.delta_seconds()`).

src/user_input.rs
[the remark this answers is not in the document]
CleanCut: Would it be clearer to name the two structs `SingleGamepadAxis` and `DualGamepadAxis`? (Drop the `Threshold` suffix)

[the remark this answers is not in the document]
zicklag: > I'm wondering if we should be managing deadzones at this level? Maybe we don't need to? If we do, we should probably use the same terminology as bevy_input.

My reasoning behind managing deadzones at this level has to do with action specific dead-zones. For instance, in my game I have a character that I use the left stick to walk around with. For controlling the character, if you tilt the stick very slightly, I still want the `Move` action to be triggered and the player to move based on how far the stick is tilting.

But I also have a separate set of actions for the Menu, such as `Next`, and `Previous` which have bindings to the left stick as well. Since `Next` and `Previous` are not variable actions, I really only want to trigger them if the stick moves more than 50% in a direction, and I want that to be configured separate from the movement deadzone.

That is a good point about the term `threshold`. That could be confusing. I'm just not sure if we can use the same terminology as Bevy because we are trying to represent a different concept. In bevy, the settings say ( roughtly ), "at which point should I ignore the stick movement". With leafwing input we are saying, "at which point should my stick movement trigger an in-game response".

Another note: since this plugin is getting our input from bevy, the `bevy_input::GamepadSettings` are still applied, but they are applied before leafwing gets the values.

> Would it be clearer to name the two structs SingleGamepadAxis and DualGamepadAxis? (Drop the Threshold suffix)

Yeah, I like that.

---

Anyway, I'll take another look at it to see if I can come up with a more intuitive way to structure the enum maybe. Or let me know if you still think that maybe we should try to get rid of local thresholds in favor of a more global setting, or if you have any other ideas. :)

CleanCut:
Note: You have to go to the `Files Changed` tab and expand the diff for `src/user_input.rs` in order to see the threads on that file. They're (annoyingly) not showing up in this `Conversation` tab of the PR like the rest of the comments. I think it's because the diff is so large. (Why is the diff thinking every line changed, anyway? Did the line endings change or something?)

src/user_input.rs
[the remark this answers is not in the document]
alice-i-cecile: I'm fine with the local thresholds :) That's not something I'd thought of before but I think it's actually the correct behavior.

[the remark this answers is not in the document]
CleanCut: > That is a good point about the term threshold. That could be confusing. I'm just not sure if we can use the same terminology as Bevy because we are trying to represent a different concept.

Mirring Bevy's terminology is probably the clearest approach. I agree there is a subtle difference, but I think the semantics are essentially the same. In both cases, input is being omitted for a certain range of values.

My suggestion would be to split your current `threshold` field into the analogous `positive_low` and `negative_low` fields from Bevy's [`AxisSettings`](https://docs.rs/bevy/latest/bevy/input/gamepad/struct.AxisSettings.html).  Then, if we want to support more settings similar to `AxisSettings`, we can similarly add the same field names to your structs.

> Or let me know if you still think that maybe we should try to get rid of local thresholds in favor of a more global setting, or if you have any other ideas. :)

Local settings are fine. I wasn't really going for global over local settings, per se. I just used Bevy's global settings as my example.  

> Anyway, I'll take another look at it to see if I can come up with a more intuitive way to structure the enum maybe. 

I wonder if it would make sense to combine `SingleGamepadAxis` and `DualGamepadAxis` into a single type. But that may be a bridge too far for this PR. I haven't been able to come up with a good design for that in my head. (Just thinking out loud on this one--I'm not sure this is actionable.)

CleanCut:
Will this PR resolve #50, or does other virtual-axis handling need to supported for that to be closed?

src/user_input.rs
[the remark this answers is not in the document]
zicklag: > My suggestion would be to split your current threshold field into the analogous positive_low and negative_low

OK, I see, that makes sense. So if you wanted to trigger on any positive axis movement, you would set the `negative_low: -1.1` to make the negative movement never trigger, and the `positive_low: 0.0` to make positive move always trigger.

I think that makes sense and handles all the real-world use-cases I can think of. Good call!

I was just messing around with alternatives and came up with this, to get rid of `threshold`, but it's overkill and I like the bevy convention better. :) 

```rust
pub enum Comparison {
    /// Value must be less than or equal to a value
    Less(f32),
    /// Value must be greater than or equal to a value
    Greater(f32),
    /// The absolute value must be less than or equal to a value
    AbsoluteLess(f32),
    /// The absolute value must be greater than or equal to a value
    AbsoluteGreater(f32),
    /// The value must be within the given range
    WithinRange(f32, f32),
}
```

> I wonder if it would make sense to combine SingleGamepadAxis and DualGamepadAxis into a single type. 

Not a bad thought, but with all the ways I've thought of it so far I think it ends up cleaner to have them separate.

We could follow up if anything comes to mind.

zicklag:
> Will this PR resolve https://github.com/Leafwing-Studios/leafwing-input-manager/issues/50, or does other virtual-axis handling need to supported for that to be closed?

I feel like it could close it, and then we open up new issues for remaining tasks?

---

I just pushed another update that covered the review points, updated some doc strings and the example, and I think this is close to being ready. Might still need more nit picking, but I think it's close.

zicklag:
OK, after trying to use the latest version I'm not totally liking how I have to watch for input in one direction of an axis:

```rust
SingleGamepadAxis {
  axis: GamepadAxisType::LeftStickX,
  negative_low: -0.5, // This triggers if the negative direction is below 0.5
  positive_low: 1.1, // This makes sure the positive direction never triggers
}
```

The `positive_low` is the confusing part, because it just needs to be any number higher than 1.0, but it isn't self-explanatory.

What do you think about using options instead?

```rust
// I want to trigger when you move the left stick left
SingleGamepadAxis {
  axis: GamepadAxisType::LeftStickX,
  negative_low: Some(-0.5), // This triggers if the negative direction is below 0.5
  positive_low: None, // This makes sure the positive direction never triggers
}
```

CleanCut:
> OK, after trying to use the latest version I'm not totally liking how I have to watch for input in one direction of an axis:

I don't like the `Option` approach here. 😕 I find the `f32` implementation more straightforward. It matches the semantics of `bevy_input` for one, which is a huge plus.

If you want to use the deadzone to silence one half of an axis, you can set it to positive or negative `1.0`--you shouldn't need to pick an arbitrary larger value. `1.0` for a total deadzone has good symmetry with `0.0` for no deadzone.

I typically wouldn't reach for deadzone settings if I wanted to ignore one half of an axis, though. I would typically handle that sort of "ignoring" in my handling code. The deadzone is typically for setting the sensitivity of when you want to start receiving input from the analog axis, not for dealing with that input it at a logical level.

If we really want to feature-creep, I'd vote for 1) the ability to map the value linearly to an arbitrary output range, and then 2) the ability to define a curve through which to transform the output non-linearly.  But let's not get that fancy in this PR. 😄

zicklag:
> If you want to use the deadzone to silence one half of an axis, you can set it to positive or negative 1.0--you shouldn't need to pick an arbitrary larger value. 1.0 for a total (half) deadzone has good symmetry with 0.0 for no deadzone.

I had to set it more than 1.0 to ignore it because I was using a `>=` check, but if we set it to a `>` check, which really wouldn't impact it functionally, that would work, and I'd be happy with that.

> I typically wouldn't reach for deadzone settings if I wanted to ignore one half of an axis, though. 

Yeah, the reason I feel the need to do it here is because leafwing is outputting "button-like" actions, that I want to be able to respond to like buttons, but that happen to be tied to an axis. Though I suppose I could switch to an opposite mentality and say that I could treat all buttons as axes by reading their `action_value()` and then I still get unified handling. :thinking: 

---

OK, I took out the options and switched the deadzone comparison to `>` and `<` instead of `>=` and `<=`. That will let you use `1.0` instead of `1.1` to ignore an axis, and I think that should work fine.

---

> If we really want to feature-creep, I'd vote for 1) the ability to map the value linearly to an arbitrary output range, and then 2) the ability to define a curve through which to transform the output non-linearly. But let's not get that fancy in this PR. smile

Definitely not for this PR, but now you've tempted me to think about it. What if we had input _middleware_. 👀 😄

Don't have time for that right now anyway, but it's a fun thought.

alice-i-cecile (approved):
I'm happy with this as an initial implementation now! I'll leave it for a day or two, so get your reviews in.

CleanCut:
@zicklag

- The implementation of `SingleGamepadAxis` looks good to me 👍🏻 
- `DualGamepadAxis` should be more similar to `SingleGamepadAxis`. We have some options here, and I'm not sure what's best:
  - Split `deadzone` out into `x_positive_low`, `x_negative_low`, `y_positive_low`, and `y_negative_low`
    - Pro: Very straightforward, Cons: lots of fields (but folks can use struct update syntax if they're okay with default values, or a helper constructor we make in the future) -- I think this is my favorite option.
  - Embed two `SingleGamepadAxis` structs inside of `DualGamepadAxis`
    - Pro: Easy to code. Con: Users will have to deal with nesting two structs inside another at definition time
  - Discard `DualGamepadAxis` and accept a tuple of `(SingleGamepadAxis, SingleGamepadAxis)`
    - Pro: Elegant solution, Con: Hard to learn that this is The Way™ to do it, when everything else has a specific type.

alice-i-cecile:
I agree with you @CleanCut; I think the first approach is the best of those options.

zicklag:
Sounds good! Pushed an update!

CleanCut:
🤔 Hmm, we can't use struct update syntax without implementing `Default`, but we can't implement `Default` without picking default `GamepadAxisType`s, which would be nonsensical. So much for using struct update syntax! 😞 Oh, well. We'll just have to lean on a nice constructor(s) in the future to improve ergonomics.

I'm going to go through the entire code again. Gimme an hour...

zicklag:
> but we can't implement Default without picking default GamepadAxisTypes, which would be nonsensical. 

We could just use the struct update syntax combined with a constructor:

```rust
DualGamepadAxis {
    x_positive_low: 0.5,
    ..DualGamepadAxis::new(GamepadAxisType::LeftStickX, GampadAxisType::LeftStickY)
}
```

Or even:

```rust
DualGamepadAxis {
    x_positive_low: 0.5,
    ..DualGamepadAxis::left_stick()
}
```

src/user_input.rs
```diff
@@ -1,444 +1,814 @@
...
+impl From<MouseButton> for UserInput {
+    fn from(input: MouseButton) -> Self {
+        UserInput::Single(InputKind::Mouse(input))
+    }
+}
+
+/// A button-like input type
+///
+/// See [`Button`] for the value-ful equivalent.
+/// Use the [`From`] or [`Into`] traits to convert from a [`InputButton`] to a [`InputMode`].
```
CleanCut, line 293:
```suggestion
/// What mode (sort of device) an [`InputKind`] originated from.
///
/// Use the [`From`] or [`Into`] traits to convert from a [`InputKind`] to a [`InputMode`].
```

src/user_input.rs
```diff
@@ -1,444 +1,814 @@
...
+                return true;
+            }
+        }
+        // If none of the inputs matched, return false
+        false
+    }
+
+    /// Is the `button` pressed?
+    #[must_use]
+    pub fn button_pressed(&self, button: InputKind) -> bool {
```
CleanCut, line 592: @alice-i-cecile `InputKind` now includes axes. Should this method be renamed (someday)? 🤷🏻‍♂️ 

```suggestion
    /// Is the `button` pressed?
    /// TODO: Rename this method since InputKind includes axes as well as buttons?
    #[must_use]
    pub fn button_pressed(&self, button: InputKind) -> bool {
```

src/user_input.rs
```diff
@@ -1,444 +1,814 @@
...
+            mouse: Some(mouse_input_stream),
+            associated_gamepad: None,
+        }
+    }
+}
+
+// Input checking
+impl<'a> InputStreams<'a> {
+    /// Is the `input` matched by the [`InputStreams`]?
+    pub fn input_pressed(&self, input: &UserInput) -> bool {
```
CleanCut, line 558: @alice-i-cecile Instead of "pressed", would it be clearer to (someday) use the term "active" or "activated"? This returns `true` for an axis if it is outside of the deadzone.

src/user_input.rs
```diff
@@ -1,444 +1,814 @@
...
+    pub fn get_input_value(&self, input: &UserInput) -> f32 {
+        let use_button_value = || {
+            if self.input_pressed(input) {
+                1.0
+            } else {
+                0.0
+            }
+        };
+
+        if let Some(gamepad) = self.associated_gamepad {
```
CleanCut, line 669: I suggest refactoring this code so that there is one `match` expression on the outside, and one or more `if` expression(s) dealing with the gamepads inside their respective `match` arm(s).  That way, you won't hit the current problem where you forget to handle non-gamepad related stuff consistently between two `if` branches.

Specifically:
- The top branch of the `if` does not handle `UserInput::Chord`, so you can only get chord values if you have a gamepad attached.
- The bottom branch of the `if` does not handle `UserInput::VirtualDPad`, so you can only get virtual direction pad values if you _don't_ have a gamepad attached.

src/user_input.rs
```diff
@@ -1,444 +1,814 @@
...
+                        let y = axes
+                            .get(GamepadAxis(gamepad, threshold.y_axis))
+                            .unwrap_or_default();
+                        Some(AxisPair::new(Vec2::new(x, y)))
+                    } else {
+                        Some(AxisPair::new(Vec2::ZERO))
+                    }
+                } else {
+                    Some(AxisPair::new(Vec2::ZERO))
+                }
```
CleanCut, line 742: Providing fake axes values is confusing.

```suggestion
                    } else {
                        None
                    }
                } else {
                    None
                }
```

CleanCut (commented):
@zicklag First, excellent work!  Thank you for providing this functionality...and just when I needed it, too!

Second, I've done a _very thorough_ review here with the intent to help improve things, so please take it in that spirit! 😄  Much of my suggestions can either be committed with a button click, or be ignored (especially wording on comments) if you like.

I did find a few items which need to be either fixed, discussed, or added to a to-do list for the future.

### VirtualDPad and UserInput

_I didn't see a logical place in the code for this discussion, so I'll do it here._

We probably need to introduce a `VirtualDPad` struct for users to pass to the `InputMap::insert*` methods.

The `UserInput` enum (and the `InputKind` enum) are _almost_ entirely hidden from the user by virtue of _nearly_ all example and documentation passing in some helper struct which implements `impl From<THING> for UserInput`.  The only "thing" that doesn't do this that I can see is the virtual direction pad. This forces the user to learn about the `UserInput::VirtualDPad` instead of using a dedicated struct like everything else.

For consistency, we should probably add a `struct VirtualDPad` and `impl From<VirtualDPad> for UserInput`.

I almost want to say we could even make the `UserInput` and `InputKind` enums private, but that might not be true, and I don't want to take the time to do that analysis right now.

alice-i-cecile:
> I almost want to say we could even make the UserInput and InputKind enums private, but that might not be true, and I don't want to take the time to do that analysis right now.

I don't think we can; it makes it much too hard to build your own abstractions on top.

src/user_input.rs
alice-i-cecile: I'm on board with that in a follow-up PR.

examples/axis_inputs.rs
```diff
@@ -0,0 +1,101 @@
+use bevy::prelude::*;
+use leafwing_input_manager::prelude::*;
+
+fn main() {
+    App::new()
+        .add_plugins(DefaultPlugins)
+        // This plugin maps inputs to an input-type agnostic action-state
```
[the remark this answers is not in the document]
alice-i-cecile: It's a list of in-game actions that correspond to behaviors. These occur at the game-logic level, and don't care about *how* they were triggered.

src/axislike.rs
```diff
@@ -1,98 +1,107 @@
...
+            AxisPair { xy: xy / magnitude }
+        }
+    }
+
+    /// Merge the state of this [`AxisPair`] with another.
+    ///
+    /// This is useful if you have multiple sticks bound to the same game action,
+    /// and you want to get their combined position.
+    pub fn merged_with(&self, other: AxisPair) -> AxisPair {
+        AxisPair::new(self.xy() + other.xy())
```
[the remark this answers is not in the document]
alice-i-cecile: Agreed with this suggestion.

src/axislike.rs
```diff
@@ -1,98 +1,107 @@
...
+///
+/// The neutral origin is always at 0, 0.
+/// When constructed; the magnitude is capped at 1, but direction is preserved.
+///
+/// This struct should store the processed form of your raw inputs in a device-agnostic fashion.
+/// Any deadzone correction, rescaling or drift-correction should be done at an earlier level.
+#[derive(Debug, Copy, Clone, PartialEq, Default, Deserialize, Serialize)]
+pub struct AxisPair {
+    xy: Vec2,
+}
```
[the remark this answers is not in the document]
alice-i-cecile: Newtyping is important here; the value range is restricted in critical ways.

src/user_input.rs
alice-i-cecile: Agreed; see #157.

src/user_input.rs
alice-i-cecile: See #157.

src/input_mocking.rs
```diff
@@ -132,6 +133,16 @@ impl<'a> MutableInputStreams<'a> {
             }
         }
 
+        if let Some(ref mut gamepad_input) = self.gamepad_axes {
+            for axis in gamepad_axes {
+                if let Some(associated_gamepad) = self.associated_gamepad {
+                    let gamepad_axis = GamepadAxis(associated_gamepad, axis);
+                    // FIXME: Allow setting axis input value
```
zicklag, line 140: I think it'd be better to follow up with this than put it into this PR. :)

zicklag (commented):
Phew! I think I covered it all. :)

I can't thank you enough for all your time reviewing @CleanCut and @alice-i-cecile. This turned out way bigger than I thought, but I'm really liking it and I think it will help a lot.

zicklag:
Whoops, I just ran into virtual DPads needing to clash with other inputs for my key binding. I'm going to see if I can fix that real quick.

src/action_state.rs
```diff
@@ -181,6 +191,43 @@ impl<A: Actionlike> ActionState<A> {
...
+    /// - Some axes, such as a variable trigger, will have a value in the range `0.0..=1.0`.
+    /// - Some buttons will also return a value in the range `0.0..=1.0`, such as analog gamepad
+    /// triggers which may be tracked as buttons or axes. Examples of these include the Xbox LT/RT
+    /// triggers and the Playstation L2/R2 triggers. See also the `axis_inputs` example in the
+    /// repository.
+    /// - Dual axis inputs will return the magnitude of its [`AxisPair`] and will be in the range
+    /// `0.0..=1.0`.
+    /// - Chord inputs will return the value of its first input.
+    ///
+    /// If multiple inputs trigger the same game action at the same time, the value of each
```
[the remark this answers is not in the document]
odecay: It may make sense in another PR to have a way to choose between[ SOCD resolution strategies.](https://www.hitboxarcade.com/blogs/support/what-is-socd)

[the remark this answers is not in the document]
zicklag: > So if you have say a move forward action bound to Dpad up and analog stick y+ and trigger both simultaneously you would expect it to return an action value of 1.0?

Yep.

> Trying to think through what the implications of this are. What is the expected behavior for an action being bound to both Dpad and analog when receiving opposite direction inputs?

It would return a value of 0.0, but the input would still register as "pressed".

zicklag:
OK, fixed DPad clashing and added a couple tests to make sure it's working.

alice-i-cecile:
@zicklag format this and I'll merge it in :) This is plenty big already; let's get this in place so others can start building off it!

zicklag:
There you go!

examples/axis_inputs.rs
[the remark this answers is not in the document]
CleanCut: Ah! So maybe something like this?

```
// This plugin maps physical inputs to logical actions
```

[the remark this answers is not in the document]
alice-i-cecile: Yep! @CleanCut I'd happily accept a PR to that effect.

ndarilek:
Figuring out if this plugin can work for my use case. I think it can, but I'm wondering if any of the following in README.md is now invalid?:

```
- The `Button` enum only includes `KeyCode`, `MouseButton` and `GamepadButtonType`.
  - This is due to object-safety limitations on the types stored in `bevy::input::Input`
  - Please file an issue if you would like something more exotic!
- No built-in support for non-button input types (e.g. gestures or analog sticks).
```

My use case is that I want to bind certain actions to swiping up/down/left/right with the right stick. The readme makes it seem like this may not be a thing yet, but this PR makes me think it might be.

Thanks.

zicklag:
Hi @ndarilek with the work-in-progress changes in the `dev` branch it is indeed possible to bind actions to directional inputs on the right stick.

This will be released with 0.5, or you can use a git dependency on the `dev` branch if you are using Bevy 0.8.

zicklag:
No problem at all! :smiley:

PR #156, opened 2022-07-09
Refactor rotation

### LINKED ISSUE #155: Why is Rotation measured "clockwise from midnight"?
@alice-i-cecile While experimenting with #151 I noticed that Rotation is measured "clockwise from midnight". Is this an intentional decision? Does it need to be this way?

It would be much more convenient to interact with standard `f32` math as well as things like `rapier2d` if Rotation were measured _counter-clockwise from the positive X axis_ (or "counter-clockwise from 3 o'clock" to put it in the same terms). 

Currently, any angle returned by `Rotation::to_radians` needs some extra calculations to get the radians value you need for any standard `f32` math. For example, I'm trying to determine the angle that a player is attempting to fly in 2D space via the axis returned by a gampad stick. Right now, after I get a value from the input manager, I've got to flip the rotation and shift the rotation origin.

From looking over the implementation of Rotation, it _seems_ like switching to the same measurement system as the rest of mathematics would be fairly straightforward. For example, the order of the first two arguments in [this line](https://github.com/Leafwing-Studios/leafwing-input-manager/blob/9a07bb13f7d77cf6efa4bf78a64d87a6aac1af53/src/orientation.rs#L331) would go back to the standard y first and then x, and we could stop [flipping the sign](https://github.com/Leafwing-Studios/leafwing-input-manager/blob/9a07bb13f7d77cf6efa4bf78a64d87a6aac1af53/src/orientation.rs#L686) in some places to restore the counter-clockwise rotation, and use the [unit X vector instead of the unit Y vector](https://github.com/Leafwing-Studios/leafwing-input-manager/blob/9a07bb13f7d77cf6efa4bf78a64d87a6aac1af53/src/orientation.rs#L692) to measure from the positive X axis.

Is this a change you would be open to? I don't want to make a PR to change this if it needs to be kept this way for some reason I don't know.

CleanCut:
- [x] I based this off of #151, so I'm going to wait for that to get merged, resolve any conflicts, and then take this out of "draft" status. At that point the diff will become much, much smaller.
- [ ] Also, I still need to go over this again and see if I missed any documentation or examples that need to be updated.
- [x] @alice-i-cecile QUESTION: #151 itself seems to be branched off of the `dev` branch. Should I change the merge target to the `dev` branch instead of the `main` branch? (Yes, done)

---

Resolves #155

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,16 @@
 # Release Notes
 
+## Version 0.?
+
+### Enhancements
+
+- Added `Direction::try_new` which never panics.
+- Converting from a `Direction` (which uses a `Vec2` of `f32`'s internally) to a `Rotation` (which uses exact decidegrees) now has special cases to ensure all eight cardinal directions result in exact degrees. For example, a unit vector pointing to the Northeast now always converts to a `Direction` with exactly 1350 decidegrees. Rounding errors may still occur when converting from arbitrary directions to the other 3592 discrete decidegrees.
+
+### Changes
+
+- Rotation is now measured from the positive x axis in a counterclockwise direction. This applies to both `Rotation` and `Direction`.
+
 ## Version 0.5
 
 ### Enhancements
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -2,6 +2,17 @@
 
 ## Version 0.4
 
+### Changes
+
+- Renamed `InputButton` to `InputKind` to reflect it's ability to represent non-button inputs such as axes.
+
+### Enhancements
+
+- Added `SingleGamepadAxis` and `DualGamepadAxis` structs that can be supplied to an `InputMap` to trigger on axis inputs.
+- Added `VirtualDPad` struct that can be supplied to an `InputMap` to trigger on four direction-representing inputs.
+- Added `ActionState::action_axis_pair()` which can return an `AxisPair` to containing the analog values of a `SingleGamepadAxis`, `DualGamepadAxis`, or `VirtualDPad`.
+- Added `ActionState::action_value()` which represents the floating point value of any action: `1.0` or `0.0` for pressed or unpressed buttons, a value in the range `-1.0..=1.0` for a single axis representing its analog input, or a value in the range `0.0..=1.0` for a dual axis representing the magnitude (length) of its vector.
+
 ### Usability
 
 - reduced required `derive_more` features
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -15,6 +15,7 @@
 - Removed the `ActionState::reasons_pressed` API.
   - This API was quite complex and not terribly useful.
   - This was not needed for axislike inputs in the end.
+- Added `Direction::try_new()` to fallibly create a new `Direction` struct (which cannot be created from the zero vector).
 
 ## Version 0.4.1
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -10,6 +10,7 @@
 ### Changes
 
 - Rotation is now measured from the positive x axis in a counterclockwise direction. This applies to both `Rotation` and `Direction`.
+  - This increases consistency with `glam` and makes trigonometry easier.
 
 ## Version 0.5
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -11,6 +11,7 @@
 
 ### Usability
 
+- If no gamepad is registered to a specific `InputMap`, inputs from any gamepad in the `Gamepads` resource will be used.
 - Removed the `ActionState::reasons_pressed` API.
   - This API was quite complex and not terribly useful.
   - This was not needed for axislike inputs in the end.
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,10 +1,6 @@
 # Release Notes
 
-## Version 0.4
-
-### Changes
-
-- Renamed `InputButton` to `InputKind` to reflect it's ability to represent non-button inputs such as axes.
+## Version 0.5
 
 ### Enhancements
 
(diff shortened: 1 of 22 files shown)


alice-i-cecile:
Yes please, target dev. This is to avoid creating misleading examples in the default branch on GitHub.

CleanCut:
> Yes please, target dev. This is to avoid creating misleading examples in the default branch on GitHub.

✅ Rebased.

I still intend to go over this again and polish it a bit more. Hopefully within the next couple days, but I'll go ahead and take this out of draft status.

CleanCut:
@alice-i-cecile You're welcome to push commits directly to this PR! Sorry for the delay in cleaning this up. I'm traveling for my day job and having a hard time to find the time to get to this!  If I'm able to find time before you do, I'll fix this up, but otherwise feel free to take it over!

alice-i-cecile:
Excellent :D Plan right now is finish #173, merge this, split the primitives into `leafwing_input_playback`, publish a 0.0 of that and then depend on it from this crate.

CleanCut:
Hmm. The UI claims conflict with `dev`, but I merged `dev` into this branch just fine.  🤷🏻‍♂️

CleanCut:
Oh...I had an out-of-date version of `dev` locally. 🤦🏻‍♂️

CleanCut:
Looks like I have some work to do to adapt to your rapid changes! 🙀

src/axislike.rs
```diff
@@ -67,15 +67,11 @@ impl AxisPair {
 
     /// The [`Direction`] that this axis is pointing towards, if any
     ///
-    /// If the axis is neutral (x,y) = (0,0), a (0, 0) `None` will be returned
+    /// See [`Direction::new`] for more details.
     #[must_use]
     #[inline]
     pub fn direction(&self) -> Option<Direction> {
-        // TODO: replace this quick-n-dirty hack once Direction::new no longer panics
-        if self.xy.length() > 0.00001 {
```
alice-i-cecile, line 75: I added my own fix for this in #176 :)

CleanCut: 👌🏻

src/orientation.rs
```diff
@@ -173,9 +175,9 @@ mod rotation_direction {
     /// use leafwing_input_manager::orientation::{Orientation, Rotation, RotationDirection};
     ///
     /// assert_eq!(Rotation::NORTH.rotation_direction(Rotation::NORTH), RotationDirection::Clockwise);
-    /// assert_eq!(Rotation::NORTH.rotation_direction(Rotation::EAST), RotationDirection::Clockwise);
-    /// assert_eq!(Rotation::NORTH.rotation_direction(Rotation::WEST), RotationDirection::CounterClockwise);
-    /// assert_eq!(Rotation::NORTH.rotation_direction(Rotation::SOUTH), RotationDirection::Clockwise);
+    /// //assert_eq!(Rotation::NORTH.rotation_direction(Rotation::EAST), RotationDirection::Clockwise);
```
alice-i-cecile, line 178: These should be fixed before merging.

src/orientation.rs
```diff
@@ -224,39 +226,40 @@ mod rotation {
     use bevy_math::Vec2;
     use core::ops::{Add, AddAssign, Div, Mul, Neg, Sub, SubAssign};
     use derive_more::Display;
+    use std::f32::consts::TAU;
 
     /// A discretized 2-dimensional rotation
     ///
-    /// Internally, these are stored in normalized tenths of a degree, and so can be cleanly added and reversed
-    /// without accumulating error.
+    /// Internally, these are stored in normalized tenths of a degree, and so can be cleanly added
```
alice-i-cecile, line 231:
```suggestion
    /// Internally, these are stored in tenths of a degree, and so can be cleanly added
```

src/orientation.rs
CleanCut: Good catch! I totally forgot about that.

src/orientation.rs
```diff
@@ -672,13 +681,7 @@ mod conversions {
...
 
         fn try_from(vec2: Vec2) -> Result<Direction, NearlySingularConversion> {
-            if vec2.length_squared() == 0.0 {
-                Err(NearlySingularConversion)
-            } else {
-                Ok(Direction {
-                    unit_vector: vec2.normalize(),
-                })
-            }
+            Direction::try_new(vec2).ok_or(NearlySingularConversion)
```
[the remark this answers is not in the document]
CleanCut: I changed `Direction::try_new` to return `Option<Direction>` to be more similar to the behavior of `Vec2::try_normalize`, which was already doing a superset of error handling and returning an `Option::Vec2`.

Then I discovered these `TryFrom` impls. I wanted to change them to use `Option` as well...but then they wouldn't implement the trait. There's a few options for moving forward:

- Leave `Direction::try_new` returning an `Option` and convert to an error in the `TryFrom` impls. (This is what we're currently doing in this PR at this moment).
- We could change `Direction::try_new` to returning a `Result`, then we'd be dealing with `Results` in all 3-ish places.
- We could delete the `TryFrom` impls and implement some custom `try_new` methods for everything and use `Option`s in all 3-ish place.

Or something else. What's your preference?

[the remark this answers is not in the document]
alice-i-cecile: Preference is for returning `Result` from everything, and deleting `try_new`. Do link the `try_from` trait in the `Direction::new` docs though, that's hard to discover!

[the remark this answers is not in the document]
alice-i-cecile: Errors are more semantically correct here; it's "something went wrong". I've also found them to be a little easier to work with when trying to compose errors from your dependencies.

RELEASES.md
```diff
@@ -1,5 +1,16 @@
...
+## Version 0.?
+
+### Enhancements
+
+- Added `Direction::try_new` which never panics.
+- Converting from a `Direction` (which uses a `Vec2` of `f32`'s internally) to a `Rotation` (which uses exact decidegrees) now has special cases to ensure all eight cardinal directions result in exact degrees. For example, a unit vector pointing to the Northeast now always converts to a `Direction` with exactly 1350 decidegrees. Rounding errors may still occur when converting from arbitrary directions to the other 3592 discrete decidegrees.
+
+### Changes
+
+- Rotation is now measured from the positive x axis in a counterclockwise direction. This applies to both `Rotation` and `Direction`.
```
alice-i-cecile, line 12:
```suggestion
- Rotation is now measured from the positive x axis in a counterclockwise direction. This applies to both `Rotation` and `Direction`.
  - This increases consistency with `glam` and makes trigonometry easier.
```

src/orientation.rs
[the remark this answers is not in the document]
CleanCut: 👍🏻 Changing...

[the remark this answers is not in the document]
CleanCut: Sorry for the delay (meetings).

✅ Done

alice-i-cecile:
Merging #177; Git wouldn't let me push to your branch and there was a tiny bit of cleanup to do.

CleanCut:
> Merging https://github.com/Leafwing-Studios/leafwing-input-manager/pull/177; Git wouldn't let me push to your branch and there was a tiny bit of cleanup to do.

Weird! I had the "maintainer can push" or whatever thing selected. 🤷🏻‍♂️

PR #165, opened 2022-07-11
Add builder methods for new axislike input types

### LINKED ISSUE #159: Add helper methods for VirtualDpad construction

Virtual dpads are commonly constructed from: mouse wheel input, left joystick, right joystick, dpads, gamepad face buttons, WASD and arrow keys.

## Expectation

There should be helper methods to easily specify these patterns.

alice-i-cecile:
Fixes #159.

I've left these relatively minimal to start; the explicit construction isn't too bad.

# Notes to reviewers

I did a bit of shuffling; the user input file was too large and these made sense in `axislike.rs`.

I don't love that users can construct / mutate these types into weird or bad states still. I think that's best left for another PR though, it's not trivial. See #166.

### DIFF
--- a/src/axislike.rs
+++ b/src/axislike.rs
@@ -1,6 +1,7 @@
 //! Tools for working with directional axis-like user inputs (gamesticks, D-Pads and emulated equvalents)
 
 use crate::orientation::{Direction, Rotation};
+use crate::user_input::InputKind;
 use bevy_math::Vec2;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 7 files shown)


zicklag (commented):
I like it! I did leave one question.

src/axislike.rs
```diff
@@ -106,3 +112,236 @@ impl AxisPair {
...
+            left: InputKind::GamepadButton(GamepadButtonType::West),
+            right: InputKind::GamepadButton(GamepadButtonType::East),
+        }
+    }
+
+    /// Constructs a [`VirtualDpad`] corresponding to the left analogue stick of a gamepad
+    ///
+    /// A small, symettric deadzone is added for convenience.
+    /// If you wish to control the dead zone yourself, construct this struct manually.
+    pub const fn left_stick() -> VirtualDPad {
```
zicklag, line 314: I love the new `arrow_keys`, `wasd`, `dpad`, and `gamepad_face_buttons` helpers, but I'm a little confused about the `left_stick`, `right_stick`, and `from_dual_gamepad_axis` helpers.

Why would you use a `VirtualDPad` with a dual gamepad axis inside it instead of just using a `DualGamepadAxis`?

It seems like it might be confusing to a user who notices that they could either use `VirtualDPad::left_stick()` or `DualGamepadAxis::left_stick()`.

The only practical difference I can think of is that a `DualGamepadAxis` can be put in a `Chord` or combined with other axes in a `VirtualDPad`, while a `VirtualDPad` can't be used in chords or combined with other inputs.

alice-i-cecile: Oh I see! This is my fault for not quite understanding the API correctly. Let me fix that and leave a clarifying doc comment.

zicklag (approved):
Looks good to me!

PR #167, opened 2022-07-12
Remove ActionState::reasons_pressed API

alice-i-cecile:
This was originally intended to be used as part of #151, but turned out not to be needed.

While technically useful for users, it involved significant complexity and computational overhead.


alice-i-cecile:
CI please...

PR #172, opened 2022-07-12
Fix broken dependencies induced by merge conflict resolution

alice-i-cecile:
Git is a reliable and intuitive piece of software.


src/input_map.rs
```diff
@@ -632,6 +632,7 @@ mod tests {
             gamepads: None,
             keyboard: Some(&keyboard_input_stream),
             mouse: Some(&mouse_input_stream),
+            mouse_wheel: None,
```
alice-i-cecile, line 635:
```suggestion
```

PR #173, opened 2022-07-13
Support mouse wheel inputs

### LINKED ISSUE #75: Support mouse wheel inputs
Bevy's [MouseButton](https://docs.rs/bevy/latest/bevy/input/mouse/enum.MouseButton.html) enum does not support mouse wheels.

IMO these can probably just be modelled as virtual buttons.

alice-i-cecile:
# Overview

Adds support for mouse wheel input types, and fixes #75.

Mouse wheel inputs are modelled as a form of `SingleAxis` / `DualAxis` user inputs. This means that lateral mouse wheels are supported by default!

If you want to extract the values, use the `input_value` family of methods. If you just want to verify that it's pressed, use the `pressed` APIs.

## Notes for reviewers

1. Sorry about the long commit history; this code / problem was very gnarly and I needed to do clean up work as I went to make this feasible to reason about.
2. I've refactored `DualAxis` to store two `SingleAxis` types, as previously suggested. This made the code significantly nicer!
3. `AxisPair` is now `DualAxisData`.
4. `InputMode` is yeet; it was quite useless in practice and wasn't playing nice with cursed mousewheel x gamepad dual axis inputs.
5. I've had to remove clamping from `DualAxisData` and elsewhere in the API, as we now use it store mouse wheel data, which isn't normalized.

## To Do

- [ ] tests, tests, tests
- [ ] add example
- [ ] verify that this is ergonomic to use when considering `ScrollUp` and `ScrollDown` as button-like
- [ ] create an issue to consider handling `ScrollUnit`  for mouse wheel events


zicklag:
I'll try to give this at least a quick once over soon. For anybody else's reference who's reviewing, here's a more tame version of the diff that compares this branch to a newer commit after the initial pass in #151.

https://github.com/Leafwing-Studios/leafwing-input-manager/compare/e8bd6e0...mouse-wheel

alice-i-cecile:
Great :) Yeah, I'm mostly interested in conceptual / API feedback; I can write tests to make sure it actually works 😂

alice-i-cecile:
Merging in now; this PR is already far too large. Fixing the input mocking should be done in a follow-up per #178.

PR #180, opened 2022-07-15
Support Single Axis Inputs in Virtual DPads

zicklag:
This makes sure that SingleAxis inputs will report a value of 0.0 if the
axis value is not withing the triggering zone. This makes it possible to
create a virtual gamepad made up of multiple single-axis inputs on the
same stick.

I ran into a scenario where in my game input mapping screen I want to map inputs to player actions such as move up, move down, etc., in a table like most game input screens are laid out. This scenario meant that it's easier to structure the input binding as a virtual DPad, even when the player is inputting analog stick directions, and they could technically put in different axes for different directions.

This commit makes the Virtual DPad output a proper axis pair data even when it contains single axis inputs for some or all of the directions. Before this commit you would end up getting a value of -2.0 when you would move left or down on the stick, and a value of 0.0 when moving the stick right or up, because the `input_value()` function was returning a value for both left and right stick movement, even though only one of them should be triggering at a time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,8 @@ serde = {version = "1.0", features = ["derive"]}
 
 [dev-dependencies]
 bevy = {git = "https://github.com/bevyengine/bevy", default-features = false, features = ["bevy_gilrs", "bevy_sprite", "bevy_text", "bevy_ui", "bevy_render", "bevy_core_pipeline", "x11"]}
-bevy_egui = { version="0.14", default-features = false }
+# Uncomment once Bevy 0.8 is relese and bevy_egui supports it
+#bevy_egui = "0.14.0"
 
 [lib]
 name = "leafwing_input_manager"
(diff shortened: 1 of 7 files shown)


alice-i-cecile:
The changes make sense to me. Good eye, and thanks for the fix!

zicklag:
No problem! It's worth noting that we might want a similar change to dual axes where they won't report a value unless the action is actually triggered.

---

Also, CI doesn't want to work because of `bevy_egui` not being compatible with bevy 0.8. Should we just comment out that dependency and example for now?

zicklag:
> Good call. Can you make that change here?

Yep. :+1;

zicklag:
OK, I pushed an update. A couple other changes I threw in there:

- Made the axis input example slightly more readable
- Fixed a bug where the `symmetric` constructor would always trigger because the negative_low was being set to the same as positive_low
- Made sure to return Some() for axis dual axis inputs, even when they aren't triggered, so that you don't get a panic trying to unwrap the axis value on an axis input.

alice-i-cecile:
Many thanks!

alice-i-cecile (approved):

PR #185, opened 2022-07-18
Control available input types using cargo features

### LINKED ISSUE #183: Use feature flags rather than optional resources
Following the discussion in https://github.com/bevyengine/bevy/issues/5302#issuecomment-1187501766, I think the correct pattern here is actually to expose feature flags, rather than optional resources (and eventreaders).

This allows us to fully cut down compile times / sizes, and correctly panic (rather than failing silently) when the app is misconfigured.

Planned split:
- Gamepad
- Keyboard
- Mouse
- Touch

All of these will be enabled by default.

alice-i-cecile:
Fixes #183. Well, attempts to.

There are some gnarly challenges here:

1. We need to run more tests in CI, one for each input feature enabled in isolation.
2. Returned tuple types (like in `UserInput::raw_inputs`) cannot have their values ignored using the cfg flags.
3. If you ignore too many fields on `InputStreams`, you get unused lifetime errors.
4. We need to make sure that tests and examples are all using the correct features.

Overall, I think this is a decent direction, but it's a hell of a yak shave. Like the label says, if you want to try your hand at this, feel free!

Cc: @mockersf (who may have good suggestions, or choose to open the closed `Option<EventReader> issue), who suggested this direction to me 😅

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,16 @@ opt-level = 3
 members = ["./", "tools/ci", "macros"]
 
 [features]
-default = ['ui']
+default = ['ui', 'gamepad', 'keyboard', 'mouse', 'touch']
+# Enables gamepad inputs
+gamepad = []
+# Enable keyboard inputs
+keyboard = []
+# Enable mouse inputs
+mouse = []
+# Enable touch inputs (not yet implemented)
+touch = []
+# Add support for [`bevy_ui`] integration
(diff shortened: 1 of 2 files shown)


alice-i-cecile:
Closing this out but feel free to remake it; fixing these merge conflicts will be harder than just starting from scratch.

alice-i-cecile:
Closing out: this will be much easier to redo.

Issue #121, opened 2022-04-17
Add performance benchmarks


alice-i-cecile:
We need to be careful to measure both latency and throughput in a relatively realistic fashion.

PR #189, opened 2022-07-27
Initial attempt at `timing` feature flag

alice-i-cecile:
## What was the problem?

Timing information is relatively niche and likely comes with a significant performance overhead (and some annoying API warts).

Users who don't care (or want to roll their own solution) should be able to opt-out.

## How did you fix it?

Hide the timing / duration pressed information and computation behind a feature flag.

##  Status

- [ ] add benchmarks to verify that total performance costs of input management is not completely negligible (#121)
- [ ] finish the migration to a feature flag by getting the code to compile
- [ ] use the benchmarks to verify a real improvement

Shelving it for now; complexity:benefit is too high for me to dedicate time to, but I figured I'll put this up here in case you, dear reader, want to tackle this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,12 @@ opt-level = 3
 members = ["./", "tools/ci", "macros"]
 
 [features]
-default = ['ui']
+default = ['ui', 'timing']
+# bevy_ui integration
 ui = ['bevy/bevy_ui']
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
Closing for now; this will be easier to do fresh.

PR #198, opened 2022-08-01
Fix Compile Without UI Feature

zicklag:
I tried to compile without default features and it failed because of a missing feature check. This fixes it.


alice-i-cecile:
Good find, thank you!

PR #215, opened 2022-08-02
Don't print a message for every clash

Pulau-Komodo:
Every action clash currently prints a message. This seems unintended.


alice-i-cecile (approved):

alice-i-cecile:
Looks good to me! I'll merge this and cut a minor version tonight.

PR #221, opened 2022-08-04
Make `DualAxis::right_stick()` use the correct y-axis

Pulau-Komodo:
It was incorrectly using the y-axis from the left stick.


alice-i-cecile:
Thank you very much!

Issue #227, opened 2022-08-08
ToggleActions cannot be constructed manually


alice-i-cecile:
> [2:47 AM] Shatur: It looks like we forgot to add a constructor for https://docs.rs/leafwing-input-manager/latest/leafwing_input_manager/plugin/struct.ToggleActions.html
> How about to have ToggleActions::new() and ToggleActions::disabled()?
> [ToggleActions in leafwing_input_manager::plugin - Rust](https://docs.rs/leafwing-input-manager/latest/leafwing_input_manager/plugin/struct.ToggleActions.html)
> Controls whether or not the `ActionState` / `InputMap` pairs of type `A` are active
> [5:18 AM] Alice 🌹: Yes please, and also make the phantom data field pub

Reviewing this, it can actually currently be constructed via `Default`, which means this isn't a critical bug.

To resolve this, I think we should:

1. Make the `PhantomData` field pub.
2. Add two associated constants, `ToggleActions::ENABLED` and `ToggleActions::DISABLED`.

alice-i-cecile:
Thanks to @Shatur for spotting this.

PR #228, opened 2022-08-08
Improve ToggleActions

Shatur:
I added the following as was suggested in #227:

- Make `_phantom` field public and rename into `phantom`.
- Add `ToggleActions::new`.
- Add `ToggleActions::ENABLED` and `ToggleActions::DISABLED`. Do we really need them? :thinking:

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,14 @@
 # Release Notes
 
+## Unreleased
+
+### Enhancements
+
+- Improve `ToggleActions`.
+  - Make `_phantom` field public and rename into `phantom`.
+  - Add `ToggleActions::new`.
+  - Add `ToggleActions::ENABLED` and `ToggleActions::DISABLED`.
+
 ## Version 0.5
 
 ### Enhancements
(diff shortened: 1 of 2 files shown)


alice-i-cecile:
I definitely see your point about redundancy. IMO the `new` method is probably the one to cut; it seems very unlikely that you're going to want to set this dynamically and I think the consts are more readable.

If you really need to set it dynamically you can now explicitly construct the type.

Shatur:
> I definitely see your point about redundancy. IMO the new method is probably the one to cut; it seems very unlikely that you're going to want to set this dynamically and I think the consts are more readable.

I thought about it, but then constants became to look like this:
```
pub const ENABLED: ToggleActions::<A> = ToggleActions::<A>{ enabled: true, phantom: PhantomData::<A> };
```
Since users never want to customize `phantom` - I assumed that with `new` it looks a little nicer. What do you think? Keep everything? Or remove `new`?

alice-i-cecile:
I like the explicitness of that. Let's remove `new`.

alice-i-cecile (approved):

PR #230, opened 2022-08-12
Add methods to `SingleAxis` and `DualAxis` to set deadzones

Pulau-Komodo:
This adds `with_deadzone` methods to `SingleAxis` and `DualAxis`, which consume and return the self with the deadzones changed.

I wanted a more ergonomic way to set deadzones. This approach is still pretty primitive compared to using the constructor, since it is limited to symmetrical deadzones. Additionally, the library itself is limited to square deadzones, But the addition is light-weight and consistent with Bevy's style, and setting deadzones primitively seems like a common enough action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         toolchain: [stable]
-        os: [windows-latest]
+        os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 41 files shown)


alice-i-cecile:
Can you retarget this to `dev`?

alice-i-cecile (approved):

alice-i-cecile:
Going to remake this quick to avoid fighting git :) Thanks!

PR #231, opened 2022-08-12
Check Axis Threshold When Detecting button_pressed

zicklag:
Checks axis thresholds when checking when an action is pressed. This must have gotten lost in a refactor, but it was a simple fix.

@ndarilek this should fix your issue in #218, and it fixed a similar issue I was having in my game.


zicklag:
Nevermind, I just realized something is up with this, because something else is wrong with the fact that `input_axis_pair()` shouldn't be returning a value here.

zicklag:
Closing in favor of a more correct fix.

zicklag:
There, fixed in #232

PR #232, opened 2022-08-12
Make Sure Gampad Axis Value Is Within Deadzone

zicklag:
The gamepad axis type wasn't making sure the returned value was within the range of the axis thresholds.

### DIFF
--- a/src/action_state.rs
+++ b/src/action_state.rs
@@ -522,7 +522,7 @@ pub struct ActionStateDriver<A: Actionlike> {
 ///
 /// This struct is principally used as a field on [`ActionData`],
 /// which itself lives inside an [`ActionState`].
-#[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Default, Serialize, Deserialize)]
 pub struct Timing {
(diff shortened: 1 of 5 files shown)


src/action_state.rs
```diff
@@ -522,7 +522,7 @@ pub struct ActionStateDriver<A: Actionlike> {
 ///
 /// This struct is principally used as a field on [`ActionData`],
 /// which itself lives inside an [`ActionState`].
-#[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Default, Serialize, Deserialize)]
```
zicklag, line 525: @alice-i-cecile I threw a couple `Eq` derives in here to satisfy clippy, do you know if that's a SemVer breaking change? I don't think it is, but I'm not sure if there's a scenario I'm not thinking of.

alice-i-cecile: Implementing new traits isn't semver breaking (which is why we can't have negative trait bounds). All good here!

zicklag:
Pushed an extra commit that updates the changelog and bumps the crate version. Let me know if it needs anything else!

alice-i-cecile (approved):

alice-i-cecile:
All good, thanks!

PR #233, opened 2022-08-12
Change CI workflow to run on ubuntu-latest

### LINKED ISSUE #196: Run CI on Linux instead of Windows
Uses less minutes, maybe faster?

TimJentzsch:
Closes #196.

I opted for `ubuntu-latest`, because it was already used in some other CI jobs.

Apparently, VS Code also auto-formatted the YAML file, I can revert this if you want.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,9 @@ jobs:
           toolchain: stable
           components: rustfmt, clippy
           override: true
+      - name: Install alsa and udev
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

TimJentzsch:
@alice-i-cecile Because the OS is matrix encoded, it's part of the job name. So the windows job is still required to complete even though it doesn't exist anymore

PR #234, opened 2022-08-12
Run clippy on examples in CI

### LINKED ISSUE #201: Ensure that all examples run in CI
They should run for a couple milliseconds each and then be closed. We can't verify correctness, but we can avoid runtime panics and compile failures.

TimJentzsch:
Closes #201.

Note that this doesn't _run_ the examples, but I think running clippy on them gives the best ROI.

Actually running may have several disadvantages:

- It's hard to set up, you'd probably need to get a list of all examples somehow.
- It could reduce CI times substantially, depending on the timeout.
- It is flaky, sometimes it might catch an error and sometimes not.

Also note that this way of setting up CI has the disadvantage that it's not run in parallel.
This might not scale well, especially with the examples now.

### DIFF
--- a/tools/ci/src/main.rs
+++ b/tools/ci/src/main.rs
@@ -16,4 +16,9 @@ fn main() {
     cmd!("cargo clippy --workspace --all-features -- -D warnings -A clippy::type_complexity")
         .run()
         .expect("Please fix clippy errors in output above.");
+
+    // Check the examples with clippy
+    cmd!("cargo clippy --examples -D warnings -A clippy::type_complexity")
+        .run()
+        .expect("Please fix clippy errors in the examples.");
 }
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
I think this is a good starting place.

I'm definitely worried about runtime panics getting missed (especially missing resources), but this is a sizable improvement and we can revisit it later.

alice-i-cecile (approved):

alice-i-cecile:
We should disable type complexity for this entire check.

TimJentzsch:
> I think this is a good starting place.
> 
> I'm definitely worried about runtime panics getting missed (especially missing resources), but this is a sizable improvement and we can revisit it later.

I wonder if we can somehow detect those paper cuts at compile time in the future

TimJentzsch:
> We should disable type complexity for this entire check.

Yea I think I forgot a `--` in the command

alice-i-cecile (approved):

PR #236, opened 2022-08-13
Fix clippy warnings to also derive `Eq`

TimJentzsch:
Clippy now emits warnings when you derive `PartialEq`, but not `Eq` when it is possible.
In the `cargo run -p ci` script, those warnings become errors.

This PR fixes these warnings and must be merged before #234 and #235.


alice-i-cecile:
Can you add this to the changelog? I'll merge it right after.

TimJentzsch:
> Can you add this to the changelog for 0.6? I'll merge it right after.

Sure, in which section do you want it? Enhancement?

alice-i-cecile:
Usability please :)

alice-i-cecile (approved):

PR #241, opened 2022-08-23
Make `UserInput` generic over the maximum number of chord inputs

### LINKED ISSUE #21: Allow users to configure maximum size of chords

TimJentzsch:
Closes #21.

`UserInput` is now generic over the maximum number of inputs that fit in a chord. The default is still `8`.

Furthermore, `UserInput::chord` doesn't panic anymore if the iterator contains more elements than the maximum size; instead, the rest of the inputs are ignored (**breaking change**).

Additionally, a `UserInput::try_make_chord` function does the same thing, but returns a `Result` and fails when too many inputs were provided. Not sure if this is the best name for this function :P

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -11,6 +11,13 @@
 ### Usability
 
 - Implemented `Eq` for `Timing` and `InputMap`.
+- `UserInput` is now generic over the maximum size of inputs for a `Chord` (the default is still `8`).
+  - If an iterator with too many inputs is passed to `UserInput::chord`, the rest of the inputs are now ignored instead of panicing.
+  - If you want to react to this case, the new `UserInput::try_make_chord` function returns a `Result` instead.
+
+### Breaking changes
+
+- `UserInput::chord` does not panic anymore if you provide an iterator with more inputs than the maximum allowed number. Instead, the rest of the elements are silently ignored.
 
 ## Version 0.5
 
--- a/src/user_input.rs
+++ b/src/user_input.rs
@@ -11,6 +11,10 @@ use crate::{
     buttonlike::{MouseMotionDirection, MouseWheelDirection},
 };
 
+/// An [`UserInput::Chord`] could not be created because there were too many input elements
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct ChordCreationError;
+
 /// Some combination of user input, which may cross [`Input`]-mode boundaries
 ///
 /// Suitable for use in an [`InputMap`](crate::input_map::InputMap)
@@ -33,6 +37,7 @@ impl<const CHORD_MAX_SIZE: usize> UserInput<CHORD_MAX_SIZE> {
     /// If `inputs` has a length of 1, a [`UserInput::Single`] variant will be returned instead.
     ///
     /// If `inputs` contains more than `CHORD_MAX_SIZE` elements, the rest of the elements are ignored.
+    /// If you want to react to this case with a [`Result`], use [`UserInput::try_make_chord`] instead.
     pub fn chord(inputs: impl IntoIterator<Item = impl Into<InputKind>>) -> Self {
         // We can't just check the length unless we add an ExactSizeIterator bound :(
         let mut length: usize = 0;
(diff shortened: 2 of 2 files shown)


TimJentzsch:
Hmm, it looks like defaults for const generics are only allowed for `structs` but not for the method implementations. So you actually need to always specify the constant for `UserInput::chord`, making it a lot more breaking of a change.

src/user_input.rs
```diff
@@ -11,32 +11,41 @@ use crate::{
     buttonlike::{MouseMotionDirection, MouseWheelDirection},
 };
 
+/// An [`UserInput::Chord`] could not be created because there were too many input elements
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct ChordCreationError;
```
alice-i-cecile, line 16: This should probably return the passed in iterator to make error recovery more feasible.

src/user_input.rs
```diff
@@ -47,6 +56,38 @@ impl UserInput {
...
+        for button in inputs.into_iter().take(CHORD_MAX_SIZE) {
+            length += 1;
+
+            if length >= CHORD_MAX_SIZE {
+                return Err(ChordCreationError);
+            }
+            set.insert(button.into());
+        }
+
+        let user_input = match length {
```
alice-i-cecile, line 83: This matching behavior is inconsistent with the behavior in the infallible version. It's also surprising, as the user has explicitly tried to make a chord.

Might still be correct, but if so we'll want a different name.

src/user_input.rs
```diff
@@ -47,6 +56,38 @@ impl UserInput {
         }
     }
 
+    /// Creates a [`UserInput::Chord`] from an iterator of [`InputKind`]s
+    ///
+    /// If `inputs` has a length of 1, a [`UserInput::Single`] variant will be returned instead.
+    ///
+    /// If `inputs` contains more than `CHORD_MAX_SIZE` elements, a [`ChordCreationError`] is returned.
+    /// If you want to silently ignore the rest of the elements, use the [`UserInput::chord`] function instead.
+    pub fn try_make_chord(
```
alice-i-cecile, line 65: Maybe we can use something super generic here? Like `new` or `try_create` or something?

This feels like it should be the default method for a lot of things.

RELEASES.md
```diff
@@ -11,6 +11,13 @@
 ### Usability
 
 - Implemented `Eq` for `Timing` and `InputMap`.
+- `UserInput` is now generic over the maximum size of inputs for a `Chord` (the default is still `8`).
```
alice-i-cecile, line 14: IMO we should probably lower the default to something much smaller, like 4, now that it can be configured, to save space for the 99% of use cases that don't want more.

src/user_input.rs
TimJentzsch: Hmm, that does make sense, but at the point of error the iterator will already be consumed for the most part, so I'm not sure how useful it will be for recovery

src/user_input.rs
alice-i-cecile: Oh, I see. Never mind!

alice-i-cecile:
> Hmm, it looks like defaults for const generics are only allowed for structs but not for the method implementations. So you actually need to always specify the constant for UserInput::chord, making it a lot more breaking of a change.

This is quite annoying and I think blocks this PR. Is there an issue for this in the `rust-lang` repo?

TimJentzsch:
> This is quite annoying and I think blocks this PR. Is there an issue for this in the `rust-lang` repo?

Looks like it's this issue: <https://github.com/rust-lang/rust/issues/98931>.

Turns out you can do something like:

```rust
let chord: UserInput = UserInput::chord(...);
```

but not

```rust
let chord = UserInput::chord(...);
```

...which is kind of weird.
Definitely not ideal for ergonomics though, IMO.

WalterSmuts:
> Hmm, it looks like defaults for const generics are only allowed for `structs` but not for the method implementations. So you actually need to always specify the constant for `UserInput::chord`, making it a lot more breaking of a change.

If possible could you post something on [the rust-lang issue](https://github.com/rust-lang/rust/issues/98931) to explain your use-case and requirements for a default type feature. AFAICT your objections to the current situation is that it:
* Has bad ergonomics
* Forces major version bump (backwards incompatibility) where a minor version bump **could've** been possible.

Would be great if we can gather a couple of use-cases to show that this is an important feature to add.

TimJentzsch:
I think this will be obsolete if we use traits instead of enums to represent the user input, which is probably the better move in the long run.

@alice-i-cecile do we already have an issue for this change? I couldn't find one myself

alice-i-cecile:
> @alice-i-cecile do we already have an issue for this change? I couldn't find one myself

Not yet. Let me make one and close this out.

Issue #216, opened 2022-08-03
A one-dimensional version of VirtualDPad


Pulau-Komodo:
It is common in games, and other software one might use game engines for, to have two inputs that are opposed to each other, without being part of a set of four. Often in such a way that doing nothing is the most reasonable interpretation of activating both inputs. For example, rotating an overview camera clockwise and counterclockwise, turning left and right in a car, or increasing and reducing throttle on a motor on a boat.

It would make sense to me to group these two inputs into a concept like `VirtualDPad` but one-dimensional, which handles cancelling-out behaviour. This would use an existing concept to reduce boilerplate somewhat.

It should also make it easy for a pair of inputs and a `SingleAxis` to control the same action, the same way a `VirtualDPad` and `DualAxis` can.

I am currently handling this case in my own code, and it is quite easy to do so, but it seems appropriate for anything so common to be handled by the library.

PR #277, opened 2022-10-26
Add VirtualAxis

johanhelsing:
Similar to `VirtualDPad`, but for one axis only.

Closes: #216

### DIFF
--- a/src/axislike.rs
+++ b/src/axislike.rs
@@ -334,6 +334,38 @@ impl VirtualDPad {
     }
 }
 
+/// A virtual Axis that you can get an [`Axis`] from
+///
+/// Typically, you don't want to store a [`Axis`] in this type,
+/// even though it can be stored as an [`InputKind`].
+///
+/// Instead, use it directly as [`InputKind::Axis`]!
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct VirtualAxis {
+    /// The input that represents the up direction in this virtual DPad
+    pub negative: InputKind,
+    /// The input that represents the down direction in this virtual DPad
+    pub positive: InputKind,
+}
+
+impl VirtualAxis {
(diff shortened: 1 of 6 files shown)


src/axislike.rs
```diff
@@ -334,6 +334,38 @@ impl VirtualDPad {
...
+///
+/// Typically, you don't want to store a [`Axis`] in this type,
+/// even though it can be stored as an [`InputKind`].
+///
+/// Instead, use it directly as [`InputKind::Axis`]!
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct VirtualAxis {
+    /// The input that represents the up direction in this virtual DPad
+    pub negative: InputKind,
+    /// The input that represents the down direction in this virtual DPad
```
johanhelsing, line 347: Oops, made some copy paste errors here.

src/axislike.rs
```diff
@@ -334,6 +334,38 @@ impl VirtualDPad {
...
 
+/// A virtual Axis that you can get an [`Axis`] from
+///
+/// Typically, you don't want to store a [`Axis`] in this type,
+/// even though it can be stored as an [`InputKind`].
+///
+/// Instead, use it directly as [`InputKind::Axis`]!
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct VirtualAxis {
+    /// The input that represents the up direction in this virtual DPad
```
alice-i-cecile, line 345:
```suggestion
    /// The input that represents the down direction in this virtual axis
```

src/axislike.rs
alice-i-cecile, line 347:
```suggestion
    /// The input that represents the up direction in this virtual axis
```

alice-i-cecile (changes requested):
Solid work! A minor doc suggestion, and this needs to be added to the change log.

alice-i-cecile (approved):

PR #283, opened 2022-11-06
Ignore keyboard and mouse if egui uses them

Shatur:


src/systems.rs
```diff
@@ -76,19 +79,28 @@ pub fn update_action_state<A: Actionlike>(
...
-    let mouse_wheel = mouse_wheel.into_inner();
+    let keycodes = keycodes.map(|keycodes| keycodes.into_inner());
+    let mouse_buttons = mouse_buttons.map(|mouse_buttons| mouse_buttons.into_inner());
+    let mouse_wheel = mouse_wheel.map(|mouse_wheel| mouse_wheel.into_inner());
     let mouse_motion = mouse_motion.into_inner();
 
+    #[cfg(feature = "egui")]
+    let (keycodes, mouse_buttons, mouse_wheel) = {
+        let ctx = egui.ctx_mut();
+        let keycodes = keycodes.filter(|_| !ctx.wants_keyboard_input());
```
alice-i-cecile, line 90:
```suggestion
        // If egui wants to own inputs, don't also apply them to the game state
        let keycodes = keycodes.filter(|_| !ctx.wants_keyboard_input());
```

Cargo.toml
```diff
@@ -20,11 +20,13 @@ members = ["./", "tools/ci", "macros"]
 [features]
 default = ['ui']
 ui = ['bevy/bevy_ui']
+egui = ['dep:bevy_egui']
```
alice-i-cecile, line 23:
```suggestion
# If this feature is enabled, egui will have priority over actions when processing inputs
egui = ['dep:bevy_egui']
```

alice-i-cecile (changes requested):
A couple of comments, and needs a release note section :)

Cargo.toml
Shatur: Agree, worth having!

alice-i-cecile (approved):

PR #289, opened 2022-11-27
Fix egui feature

Shatur:
Unfortunately, we can't clear events and inputs - it "breaks" other plugins, such as egui because it relies on these bevy resources.

For example, without this PR after clicking on a widget egui returns `false` from `wants_pointer_input` which results in clicks through widgets.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,11 @@
 # Release Notes
 
+## Unreleased
+
+### Bugs
+
+- `egui` feature now works correctly.
+
 ## Version 0.7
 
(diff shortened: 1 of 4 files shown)


alice-i-cecile (changes requested):
Ah, makes sense! Can you add this to the release notes for the next minor version?

Shatur:
> Can you add this to the release notes for the next minor version?

Yes, just did it!
I thought of this just right before your message. :smile:

alice-i-cecile (approved):

PR #282, opened 2022-11-04
Custom serde implementation for `InputMap`

alradish:
Attemp to add custom serde implementation for `InputMap`
```Rust
InputMap(
    map: {
        Run:  [Single(Keyboard(LShift)),Single(Keyboard(RShift))],
        Jump: [Single(Keyboard(Space))],
        Hide: [Chord([Some(Keyboard(R)),Some(Keyboard(E)),None,None,None,None,None,None])]
    }
)
```

To achieve this format, I added four new traits::
- `impl<A: Actionlike> FromIterator<(A, UserInput)> for InputMap<A>`
- `impl<A: Actionlike> From<HashMap<A, Vec<UserInput>>> for InputMap<A>`
- `impl<A> Serialize for InputMap<A> where A: Actionlike + Serialize + Eq + Hash + Ord`
  - `Eq + Hash` since I'm using map to serialize inputs
  - `Ord` is for `BTreeMap` so all actions will always serialize in the same order (I mainly need this to test serialization)
- `impl<'de, A> Deserialize<'de> for InputMap<A> where A: Actionlike + Deserialize<'de> + Eq + Hash`
  - `Eq + Hash` to be able deserialize `HashMap` 

Add new crate [serde-test](https://docs.rs/serde_test/latest/serde_test/) in dev-dependencies to be able to test a custom implementation of serde

Fixes #279

PR #294, opened 2022-12-07
Custom serde implementation and `TypeUuid` for `InputMap`

alradish:
- Bring changes from #282 into main
- Add `TypeUuid` for `InputMap` to be able use it as asset without wrapper


alice-i-cecile:
Thanks a ton for doing this! Sorry for messing up my merge 😅

PR #295, opened 2022-12-10
Derive `Reflect` for `ActionState`

johanhelsing:
Makes it possible to inspect:

image

And also necessary for `ActionState` to be rolled back by bevy_ggrs.


alice-i-cecile:
Excellent!

alice-i-cecile (changes requested):
Once there's a change log note I'll merge this in.

alice-i-cecile (approved):

PR #301, opened 2023-01-15
Implement PressScheduler

TheRawMeatball:
## What was the problem?

Introduce an easy way to trigger action presses from within the update loop (e.g. a UI implemented with egui)

## How did you fix it?

Implement the optional PressScheduler helper which lets users easily schedule presses that are realized on the next frame

## \[Optional\] What needs special attention?

I added some preliminary documentation, but I'm sure it could be improved.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -7,6 +7,7 @@
 - Added custom implementation of the `Serialize` and `Deserialize` traits for `InputMap` to make the format more human readable.
 - Added `TypeUuid` for `InputMap` to be able use it as asset without wrapper
 - `ActionState` and its fields now implement `Reflect`. The type is automatically registered when the `InputManagerPlugin` is added.
+- Added `PressScheduler`.
 
 ## Version 0.7.1
 
--- a/src/press_scheduler.rs
+++ b/src/press_scheduler.rs
@@ -1,4 +1,9 @@
 //! This module contains [`PressScheduler`] and its supporting methods and impls.
+//!
+//! The [`PressScheduler`] is an optional addition to an [`InputManagerBundle`](crate::InputManagerBundle),
+//! which allows for a system that runs in [`CoreStage::Update`] later to correctly press an action.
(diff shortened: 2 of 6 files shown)


src/press_scheduler.rs
```diff
@@ -0,0 +1,39 @@
+//! This module contains [`PressScheduler`] and its supporting methods and impls.
```
alice-i-cecile, line 1: This could be more useful: readers will already see this right below in the module docs.

alice-i-cecile (changes requested):
Looks good; nice and simple. Blocking:

- add to RELEASES.md
- add a couple integration tests

Optional:

- improve module docs
- add an example (ideally with egui, motivating why you wanted this feature)
- add a comment to explain your very nice zip/map/chain shenanigans to merge components and resources

src/press_scheduler.rs
TheRawMeatball: Should I add extra detail about how the press scheduler is supposed to be used here or on the `PressScheduler` type? Ideally I wouldn't put a doc comment on a file with such few items in the first place, but, deny(missing_docs) forced my hand :p

alice-i-cecile: Hmm. Module level docs please.

RELEASES.md
```diff
@@ -7,6 +7,7 @@
 - Added custom implementation of the `Serialize` and `Deserialize` traits for `InputMap` to make the format more human readable.
 - Added `TypeUuid` for `InputMap` to be able use it as asset without wrapper
 - `ActionState` and its fields now implement `Reflect`. The type is automatically registered when the `InputManagerPlugin` is added.
+- Added `PressScheduler`.
```
alice-i-cecile, line 10:
```suggestion
- Added `PressScheduler`, used to defer action presses until the start of the next frame to ease scheduling.
```

alice-i-cecile (approved):

PR #306, opened 2023-01-23
Implement the DynAction type and co.

TheRawMeatball:
## What was the problem?

Using this crate in a complex environment where not all actions can be defined in the same place was overly complicated.

## How did you fix it?

Introduce the new `DynAction` abstraction that simplifies this.

### DIFF
--- a/src/dynamic_action.rs
+++ b/src/dynamic_action.rs
@@ -1,0 +1,122 @@
+//! The dynamic action type.
+//!
+//! This type is primarily intended for use when declaring every single action in a single enum isn't feasible,
+//! for example due to a complex crate hierarchy where various feature crates want to individually declare their own
+//! actions without needing to coordinate with other crates.
+//!
+//! Example:
+//! ```
+//! # use leafwing_input_manager::dynamic_action::{DynActionMarker, DynActionRegistry};
+//!
+//! // In crate `feature_one`
+//! #[derive(DynActionMarker)]
+//! struct FeatureOneAction;
+//!
+//! fn register_feature_one_actions(registry: &mut DynActionRegistry) {
+//!     // There could potentially be many types registered here
+//!     registry.register::<FeatureOneAction>();
+//! }
+//!
+//! // In crate `feature_two`
+//! #[derive(DynActionMarker)]
+//! struct FeatureTwoAction;
+//!
+//! fn register_feature_two_actions(registry: &mut DynActionRegistry) {
+//!     // There could potentially be many types registered here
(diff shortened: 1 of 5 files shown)


src/dynamic_action.rs
```diff
@@ -0,0 +1,142 @@
...
+use once_cell::sync::OnceCell;
+
+use crate::Actionlike;
+
+pub use leafwing_input_manager_macros::DynActionMarker;
+
+static DYN_ACTION_MAP: OnceCell<HashMap<TypeId, usize>> = OnceCell::new();
+static REGISTRY_CREATED: OnceCell<()> = OnceCell::new();
+
+/// The runtime representation of action s declared via marker types
```
alice-i-cecile, line 53:
```suggestion
/// The runtime representation of actions declared via marker types
```

src/dynamic_action.rs
```diff
@@ -0,0 +1,142 @@
...
+    prelude::{App, Resource},
+    utils::HashMap,
+};
+use once_cell::sync::OnceCell;
+
+use crate::Actionlike;
+
+pub use leafwing_input_manager_macros::DynActionMarker;
+
+static DYN_ACTION_MAP: OnceCell<HashMap<TypeId, usize>> = OnceCell::new();
```
alice-i-cecile, line 50: Comments please for this strategy: I can follow what it does but it's definitely going to be rough for new contributors.

src/dynamic_action.rs
```diff
@@ -0,0 +1,142 @@
...
+//!
+//! Example:
+//! ```
+//! # use leafwing_input_manager::dynamic_action::{DynActionMarker, DynActionRegistry};
+//!
+//! // In crate `feature_one`
+//! #[derive(DynActionMarker)]
+//! struct FeatureOneAction;
+//!
+//! fn register_feature_one_actions(registry: &mut DynActionRegistry) {
```
alice-i-cecile, line 15: Hmm, I think this may be clearer with an app extension method.

It's hard for beginners to piece together exactly how you're supposed to store and mutate the `DynActionRegistry`.

alice-i-cecile (changes requested):
Finally found the focus to review this.

Much less cursed than I expected: genuinely quite happy with it. I think we should probably try to make it a bit less scary to understand, but once that's done I'm content to merge and ship this.

TheRawMeatball:
CI failiure not linked to this PR, will wait for the fix to land on main

alice-i-cecile (approved):

Issue #242, opened 2022-08-23
Detect currently active input device


TimJentzsch:
## What problem does this solve?

As a user, I want to have UI overlays that fit the input device I'm currently using (e.g. either shortcut keys when I'm using my mouse and keyboards, or gamepad glyphs if I'm using a gamepad).

As a developer, I want an easy way to detect which device is currently active, so that I can adapt my UI accordingly.

## What solution would you like?

I think the easiest way would be to provide a `Resource` which contains an `enum` indicating the currently active device. This `enum` should be updated by the plugin automatically.

Ideally, for controllers it would even identify the device type, e.g. if it's an Xbox of PS controller, so that the glyphs can be shown accordingly. But I'm not sure if that's possible.

## [Optional] How could this be implemented?

- Init the resource with some default value and let the dev provide a default if they want.
- Update the resource in a system which listens to input events:
  - A key/button press immediately changes the resource to this device.
  - Mouse or joystick movement only changes the resource if a threshold is exceeded (because this can happen accidentally).

## [Optional] What alternatives have you considered?

This could also be handled manually by the developer. I think a built in solution that works well would be nice though. This makes it easier to add proper gamepad support.

## Related work

- The [Steamworks documentation](https://partner.steamgames.com/doc/features/steam_controller/getting_started_for_devs#5) has a guideline that on-screen icons and the cursor should match the active input device.
- The documentation also [links this article](https://www.gamedeveloper.com/design/the-5-golden-rules-of-input) which lays down more rules for game input and ideas on how to implement them. It also covers the detection of the active input device.
  - > I call this the Timestamp Input Engine. The main idea is that you store 2 timestamps that indicate the time when the last input event was received for both gamepad and keyboard/mouse. Every time you detect a valid input event, you update the appropriate timestamp variable with the current game time. Make sure to use elapsed game time, not delta-time or a time that might get paused or dilated.
  - > Now that you have timestamps indicating the last gamepad and keyboard events, the actual input detection is as easy as seeing which timestamp is greater.

TimJentzsch:
> Ideally it should be toggleable via plugin settings.

Why is this necessary? It would already be optional to use the provided resource.
Or is it to save resources by disabling the updating system?

alice-i-cecile:
The latter :) Most desktop apps don't care about this functionality, so it's good to split it out in some way.

Perhaps an entirely seperate plugin?

PR #308, opened 2023-01-31
Add InputKind information to ActionData

vabrador:
# What is the problem?
If an InputMap is configured to funnel actions from different sources (keyboard, mouse, controller) into singular Actions, information about the source of an Action is lost by the time you have a system that takes ActionState<Actionlike>. Under many circumstances this is fine, and as-intended, to maintain a clean input abstraction.

In my situation, I'm building a UI navigation library that supports switching between mouse-based UI control and gamepad- or keyboard-based UI control. I'm using leafwing-input-manager as a dependency, and I rely on querying ActionStates to receive events like moving the selection cursor, and confirming or cancelling the current selection or context.

Having some information about the InputKind(s) that triggered an Action is important because, for example, when the user switches from mouse control to gamepad control, the application has to activate a selection cursor and jump to a nearby UI element so the gamepad has something to manipulate. Since my application's entire input abstraction is leafwing-input-manager's ActionState, I want to be able to ask the ActionState for a given Actionlike (e.g. Confirm, or NavigateUp, or Cancel) what InputKind fired it, so I can ask: "Did this action involve mouse-based input?"

# What is the solution?

A quick solution is to add a field `inputs: Vec<UserInput>` to `ActionData`. It's just a copy of the inputs that were used to generate the Action in the first place. It's not intended to be a simple API; but it enables the user to still use the helpful ActionState abstraction everywhere without losing the ability to do more advanced logic later on in development.

To make answering the most useful questions ("is it mouse" / "does it have X InputKind") easier, I added `involved_mouse_input(action)` and `involved_input_kind(action, input_kind)` to `action_state.rs` by way of some additional methods in `user_input.rs`.

Downstream, consumer code can look like this:
```rust
pub fn my_system(query_input: Query<&ActionState<MenuInput>>) {
    // input = query_input.get() ...
    let is_mouse_click = input.involved_mouse_input(MenuInput::Confirm);
}
```
Or, consumers can get fancier by iterating over `input.action_data[action.index()].inputs`. For example, a game might want to detect a keyboard vs. a gamepad button press and display different tooltip button graphics.

## What are the drawbacks to this solution?

- `Vec<UserInput>` adds heap allocation to `ActionData` and could be seen as a sledgehammer-to-nail situation. Maybe we could use `PetitSet` or some other construction to keep the data more lightweight.
- Because `UserInput` doesn't implement Reflect/FromReflect, I had to annotate the field with `#[reflect(ignore)]` to get it to compile. However, I think that if this information *is OK* to be in ActionData, it should probably also support reflection; I'm assuming that reflection is required for a feature like artificial input playback, which should absolutely support methods like "involved_mouse_input()".
    - I'd be willing to look deeper into getting UserInput to have Reflect/FromReflect, but it looks like it might be a large change, so I wanted buy-in first.

## Are there alternative solutions?

- Alternative 1. Just use Bevy's raw Input system.
    - I first tried this and I realized it was not very feasible. You need to add a new very specific Input<MouseButton> SystemParam to each system that newly needs to ask "is it mouse". This is a headache for one thing, but it also adds another Input abstraction dependency to your systems, and you have to remember that those two dependencies (ActionState, Input<Mouse___>) are inter-related.
    - Using raw Bevy input prevents you from using a utility method to ask the question "was this a gamepad input of some kind?" and to ask that question of a System, you need to add up to four (4) SystemParams:
    ```
    raw_bevy_gamepads: Res<Gamepads>,
    raw_bevy_gamepad_button_inputs: Res<Input<GamepadButton>>,
    raw_bevy_gamepad_button_axes: Res<Axis<GamepadButton>>,
    raw_bevy_gamepad_axes: Res<Axis<GamepadAxis>>,
    ```
    - Using Bevy raw Input would break compatibility with artificial input.
- Alternative 2. Use some other SystemParam to get information about the InputKinds that caused an Actionlike to trigger.
    - I don't know if leafwing-input-manager provides an alternative means to get this information given just an Actionlike.

### DIFF
--- a/src/action_state.rs
+++ b/src/action_state.rs
@@ -1,6 +1,7 @@
 //! This module contains [`ActionState`] and its supporting methods and impls.
 
 use crate::Actionlike;
+use crate::user_input::{UserInput, InputKind};
 use crate::{axislike::DualAxisData, buttonlike::ButtonState};
 
 use bevy::ecs::{component::Component, entity::Entity};
@@ -35,6 +36,11 @@ pub struct ActionData {
     /// Actions that are consumed cannot be pressed again until they are explicitly released.
     /// This ensures that consumed actions are not immediately re-pressed by continued inputs.
     pub consumed: bool,
+    /// What inputs are responsible for this Action?
+    /// 
+    /// This is a plain copy of the inputs used to generate this Action. Usage may be slightly advanced, but you can use this for example to determine whether an input was generated from a keypress or from a mouse button.
+    #[reflect(ignore)]
+    pub inputs: Vec<UserInput>,
 }
 
 /// Stores the canonical input-method-agnostic representation of the inputs received
@@ -111,6 +117,7 @@ impl<A: Actionlike> ActionState<A> {
 
             self.action_data[i].axis_pair = action_data[i].axis_pair;
             self.action_data[i].value = action_data[i].value;
+            self.action_data[i].inputs = action_data[i].inputs.clone();
         }
     }
 
@@ -464,6 +471,20 @@ impl<A: Actionlike> ActionState<A> {
         A::variants()
             .filter(|a| self.just_released(a.clone()))
             .collect()
+    }
+
+    /// Was this `action` fired with any kind of mouse input?
+    #[inline]
+    // #[must_use]
+    pub fn involved_mouse_input(&self, action: A) -> bool {
+        self.action_data[action.index()].inputs.iter().any(|i| i.contains_mouse_input())
+    }
+
+    /// Was this `action` fired using some input that involved the specified `InputKind`?
+    #[inline]
+    // #[must_use]
+    pub fn involved_input_kind(&self, action: A, input_kind: InputKind) -> bool {
+        self.action_data[action.index()].inputs.iter().any(|i| i.contains(input_kind))
     }
 
     /// The [`Instant`] that the action was last pressed or released
--- a/src/input_map.rs
+++ b/src/input_map.rs
@@ -359,6 +359,9 @@ impl<A: Actionlike> InputMap<A> {
             if !inputs.is_empty() {
                 action_data[action.index()].state = ButtonState::JustPressed;
             }
+
+            // Clone the raw input data into the action_data. This is useful for inspecting what input originated a particular action if you have more than one input mapped to a single action.
+            action_data[action.index()].inputs = inputs.clone();
         }
 
         // Handle clashing inputs, possibly removing some pressed actions from the list
--- a/src/user_input.rs
+++ b/src/user_input.rs
@@ -144,6 +144,42 @@ impl UserInput {
         }
     }
 
+    /// Whether the underlying InputKind(s) contain mouse input. This includes MouseButton, MouseMotion, and MouseWheel InputKinds. For UserInput Chords, this function returns true if any InputKinds in the chord involve mouse input.
+    pub fn contains_mouse_input(&self) -> bool {
+        match self {
+            UserInput::Single(ik) => {
+                ik.is_mouse_input()
+            },
+            UserInput::Chord(set) => {
+                set.iter().any(|ik| ik.is_mouse_input())
+            },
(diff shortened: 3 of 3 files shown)


vabrador:
Makes sense. I noticed that about `reasons_pressed`, so I wanted to try to give the use-case a fair treatment.

If the extra weight or complexity on ActionState / ActionData is really problematic, I wonder if there's an alternative that exposes a non-ActionState queryable component with additional information stored for a given Action that the user can "opt in" to for a system. I guess this could even be feature-gated if the extra computation overhead or learning curve is a concern, or gated on a Resource-based config.

I guess as a design question, what is the biggest concern? ActionData size, usage of the heap, input system update complexity, or sheer API complexity to the end user?

alice-i-cecile:
> I guess as a design question, what is the biggest concern? ActionData size, usage of the heap, input system update complexity, or sheer API complexity to the end user?

Primary concerns are actually:

1. I'm strongly considering reworking `InputKind` etc to use a boxed trait object, rather than a nested enum, due to memory usage, and I'm not sure how this would fit into that design.
2. API complexity to user.
3. Runtime costs (memory and compute) for a niche feature. I similarly want to make duration_pressed optional somehow.

vabrador:
OK, I think those are reasonable concerns. I do think that, while this feature is arguably niche, I don't really see any other place but leafwing-input-manager for this input discrimination to exist, for a project that:
- reaches a certain level of complexity,
- uses leafwing-input-manager's input abstractions across many systems,
- aims to support smooth, live transitions between different input devices, or to adjust UI as the user switches to different input devices.

Even if I implemented some custom project-internal resource that tracks recent device input kinds, I would lack an important association between a _specific_ input event exposed through ActionState and the device that was responsible for that change.

So maybe there's another design approach that will meet your concerns that will also solve my issue and allow me to continue using the input manager's usefully-narrow API.

Fundamentally my project needs to ask this question: Did this Action come from a Gamepad, Keyboard, or Mouse device?
```rs
pub struct ActionData {
// ... other fields ...
    /// What kind of devices were responsible for this Action?
    /// Actions driven by input chords may cause this set to contain more than one InputDeviceKind.
    ///
    /// Also consider using the `involved_mouse_input()` and related functions.
    pub source_devices: PetitSet<InputDeviceKind, 4>
}
```
Looking at the Bevy raw input systems and the `update_action_state`, they really only focus on three kinds of devices: Gamepad, Keyboard, or Mouse device events. So we can define a new Enum that is (1) small, (2) self-descriptive, and (3) fully covers what Bevy & leafwing-input-manager *currently* support insofar as devices:
```rs
pub enum InputDeviceKind {
    Mouse,
    Keyboard,
    Gamepad
}
```
This is enough information to implement the most important function my project uses on ActionState for input discrimination, which is just, "was it mouse?" (`involved_mouse_input()`)

I think this design addresses the concerns you laid out:
> 1. I'm strongly considering reworking InputKind etc to use a boxed trait object

This approach doesn't rely on InputKind at the API level; if InputKind's backing storage changes significantly, the relevant mapping from InputKind to setting the correct InputDeviceKind is the only thing that would change with it (and, most likely, only trivially)

> 2. API complexity to user.

While adding a new Enum for one feature is _not ideal_, the Enum itself is tiny and self-explanatory. Also, this should make the reflection issue disappear.

> 3. Runtime costs (memory and compute) for a niche feature. I similarly want to make duration_pressed optional somehow.

Conditionally setting this enum during the action_state update is cheaper than copying a whole InputKind Vec, and if duration_pressed does become optional in the future, this could become optional in the same way.

To be clear I haven't implemented this yet, but if you think this design has mergeability, I can rework this PR and implement it this way instead.

vabrador:
Ah, just noting here that I think this PR would address this related issue: #242

alice-i-cecile:
> (Contrary to the proposed solution in that issue, I think having device information on Actions is better than having the latest device on a Resource because the context of an Action matters for the application to decide what to do given interaction from a new device.)

Very much agree here.

alice-i-cecile:
Closing this out; I'd like to tackle this with input processing instead.

PR #312, opened 2023-02-04
Add support for scancode keyboard input

### LINKED ISSUE #179: Support ScanCode input bindings

Bind to actions based on the scancode sent. These correspond to the semantic values of keys, not just their position on the keyboard.

As a result, these are essential to supporting non-US keyboards (and for text entry).

Add an `InputKind` variant that supports this.

## [Optional] How could this be implemented?

In order to handle this correctly, we'll need to parse the keyboard event streams themselves.

Assigning semantic value is out of scope for a first attempt; ideally this is handled at the winit level: https://github.com/rust-windowing/winit/issues/1904

## Related work

See https://github.com/bevyengine/bevy/issues/2052 and https://github.com/bevyengine/bevy/issues/862 for upstream work in Bevy.

More discussion in winit: https://github.com/rust-windowing/winit/issues/753

### LINKED ISSUE #307: Add support for scan codes

Games often require input keys that should maintain their physical location on the keyboard, regardless of which keyboard layout is used.

The most common example are WASD controls.
On the AZERTY keyboard layout (commonly used in France and Belgium), the W and A keys are at completely different positions, making the movement controls infeasible.
Instead, ZQSD should be used on this layout.

The user should have an option to specify the physical location of the key instead of the logical key output.
Ideally, both formats could be combined freely.

Most likely, `winit`'s [`ScanCode`](https://docs.rs/winit/latest/winit/event/type.ScanCode.html) is the solution here.

However, dealing with these numbers directly makes the code hard to reason about.
We most likely need a nice helper API, which for example allows you to specify the key based on a specific layout (e.g. QWERTY) and the scan code is derived from that.

## Related work

The web has [`KeyboardEvent.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) which provides the physical location of a key based on its position on the QWERTY layout.

TimJentzsch:
Closes #307, closes #179.

This PR adds support for scan code keyboard input, i.e., using the key location instead of the logical key value.
This makes sense when you care about the physical location of the key rather than which letter it represents.

This is an important feature for games, as the classical WASD movement input should have keys that are closely together, in the triangle shape.
When a player uses the French AZERTY keyboard layout instead of the US QWERTY one, the ZQSD keys make more sense (which are on the same position as the WASD keys on the QWERTY layout).

You can now convert a `ScanCode` into `InputKind` to achieve this.
Furthermore, I added a helper enum `QwertyKeyLocation`, which makes it easy to select the correct scan codes based on their keys on the QWERTY keyboard layout.

The classical virtual dpad can now be defined like this:
```rs
VirtualDPad {
    up: QwertyScanCode::W.into(),
    down: QwertyScanCode::S.into(),
    left: QwertyScanCode::A.into(),
    right: QwertyScanCode::D.into(),
},
```

## TODO

- [x] Add missing definitions for `QwertyScanCode` (mostly the Numpad keys).
- [x] Check scan codes on MacOS as they might be different. If they are, we need to create a second definition of `QwertyScanCode` which is selected based on the `target_os`.

## Feedback Wanted

- Do you have better ideas for the naming here? It's kind of confusing now that we have `InputKind::Keyboard` but also `InputKind::KeyLocation`. ~~Also `QwertyKeyLocation` can probably be improved, maybe `QwertyScanCode` makes more sense here?~~ I renamed it to `QwertyScanCode`.
- Proof reading of the values for `QwertyScanCode` is appreciated, as it will be very hard to find errors here after merging.


afonsolage (commented):
I think it would be better to add another example for physical keyboard mapping, since the virtual_dpad example aims to example how to solve a different problem.

Using the new `QueryScanCode` may be confusing for new users.

TimJentzsch:
@afonsolage I reverted the changes to the virtual_dpad example and created a new one. Please let me know if you think this works better :)

TimJentzsch:
I created a mini Bevy app to check which key has which scan code: <https://github.com/TimJentzsch/bevy_scan_code_test>

It looks like the arrow key scan codes are different, I'll have to double check why

afonsolage (approved):

janhohenheim:
Did you verify that this behaves well in WASM? I had some surprises there recently when using `ScanCode`s 😅

TimJentzsch:
Good point, I didn't test that yet.
We should do that before merging

TimJentzsch:
Looks like the values are indeed different for WASM, I found a reference here: <https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode#value_of_keycode>

There might be some differences on WASM between the OS, which kinda sucks for us since we have the same compile target on WASM for all OS.
But it looks like _most_ keys are the same so it should be fine.

Anyway, time to copy another batch of hex values... :upside_down_face:

TimJentzsch:
@janhohenheim I noticed that for Foxtrot you sampled the scan codes by hand. Do you happen to have a MacOs device with Apple keyboard available for testing?

janhohenheim:
@TimJentzsch if by Apple keyboard you mean the regular keyboard on a MacBook, sure do! How can I help?

TimJentzsch:
@janhohenheim Yes!
Would you be able to run this testing app (<https://github.com/TimJentzsch/bevy_scan_code_test>), press a bunch of keys with the US QWERTY layout active and copy the CLI logs here?
That would be very helpful to verify that I set it up correctly.

Most important would be the following:
- The WASD keys
- The arrow keys
- Tab, CapsLock, Shift, Backspace, Enter, Esc
- The "bottom row" of the keyboard (e.g. Control, Command, Space, etc.)
- The key left to the `1` key and the keys between `0` and Backspace
- Some number keys
- Some numpad keys (if you have it on the keyboard)
- Some function keys
- The key group above the arrow keys (commonly Insert, Delete, etc.)

Thank you very much!

janhohenheim:
On it. For context, this is the US keyboard on macOS:
image
Holding shift:
image
For completeness' sake, here's holding option, but you probably don't care about these.
image

janhohenheim:
Done, made a gist: https://gist.github.com/janhohenheim/76ec639ef658a73dfecef2b50c21ab6c
I cleaned up the output with a little regex for you to have better readability.
I pressed all keys from upper left to bottom right without shift, since pressing shift change neither the keycodes nor scancodes. I'm a bit surprised by this, since I thought the keycode should be different, but eh.
The F11 key is missing because pressing it makes the app lose focus by definition, so it wasn't captured 😄

TimJentzsch:
Thank you very much for this thorough review!

It actually helped me to find a couple of errors:

- The `Enter` value I had was still the one from Windows/Linux, I forgot to update it.
- The value for the `RWin` (right command key) was missing in the list I based the values on.
- My `RAlt` value was duplicated from the `RControl` value, I probably switched up the line there.

But otherwise it seems to match, so that makes me more confident in the bindings :)

TimJentzsch:
Should `QwertyScanCode` be in the prelude?

alice-i-cecile:
> Should `QwertyScanCode` be in the prelude?

Hmm. Sure!

bash:
> But I pinged @bash irl, who has an apple wireless keyboard with a numpad; maybe they can help.

I'll try to run the [bevy_scan_code_test](https://github.com/TimJentzsch/bevy_scan_code_test) thing tomorrow on my Mac keyboard with a numpad and report the results :)

TimJentzsch:
Thank you @bash! This allowed me to catch another error, this whole thing is very prone to typos...

The numpad stuff seems to be correct though, except that I named them differently, because the Windows keys are at different positions (and have a key less).

janhohenheim:
Could the `[VirtualDPad](https://docs.rs/leafwing-input-manager/latest/leafwing_input_manager/axislike/struct.VirtualDPad.html)` also support `ScanCode`s in its convenience methods, i.e. [`wasd`](https://docs.rs/leafwing-input-manager/latest/leafwing_input_manager/axislike/struct.VirtualDPad.html#method.wasd)and [`arrow_keys`](https://docs.rs/leafwing-input-manager/latest/leafwing_input_manager/axislike/struct.VirtualDPad.html#method.arrow_keys)?

alice-i-cecile:
I would like that, but I'm fine to defer it to another PR.

janhohenheim:
I'm using this branch in [Foxtrot](https://github.com/janhohenheim/foxtrot) now, but it is broken on wasm: [live demo](https://github.com/janhohenheim/foxtrot).
As you can see, the mouse movement is picked up correctly, but all keyboard inputs are ignored. This did not happen with the `KeyCode` based code. This behavior is consistent across Windows 10 and macOS using Firefox and Chrome.
The native builds are working as intended. 

For reference, this is how I setup my `InputManagerBundle`: https://github.com/janhohenheim/foxtrot/blob/main/src/player_control/actions.rs#L71

TimJentzsch:
Thanks @janhohenheim for reporting this issue!

This was caused by a dumb copy/paste error which resulted in the macOS configuration being used on WASM.
I think I fixed it now, can you check if it works on Foxtrot?

`VirtualDPad::wasd` also uses scan codes now.
I didn't change the arrow keys, because they should work the same on all keyboard layouts, but I'm open to changing them as well if needed.

janhohenheim:
Cool, I'll check in an hour 🙂 
Minor nitpick: your code calls the target "WASM", but the official capitalization is "Wasm" 😉

TimJentzsch:
I think the last remaining part is figuring out why some of the Windows/Linux scan codes appear to be wrong. At least the ones prefixed by `e0_` seem to be different than the Set 01 spec I found.
Either the scan codes are not actually from Set 01 or I need to pull them from another source.

alice-i-cecile:
Okay, code quality is solid. Going to test this out on my Windows machine and see what I get.

alice-i-cecile:
Windows data for you @TimJentzsch: https://gist.github.com/alice-i-cecile/bc86f697497a44228ab20ced8b463a3c

TimJentzsch:
The Windows keycodes seem to watch what I got.

But it seems like Linux has different bindings, at least on my machine.
For example, I'm getting:

- Left: `0x69`
- Up: `0x67`
- Right: `0x6a`
- Down: `0x6c`

But I couldn't find this combination on the specs I found.
I also checked <http://www.quadibloc.com/comp/scan.htm> which was used for Foxtrot, but it doesn't seem to match the codes that I get.

janhohenheim:
@TimJentzsch I blindly copied values for Linux without testing them, so don't trust those too much 😉

TimJentzsch:
I think I found the Linux values through [this winit PR](https://github.com/rust-windowing/winit/pull/1932), it links to the Linux source directly: <https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h>

From the few values I checked it seems to match up.

janhohenheim:
That has to be the most trustworthy source you could find 😄

alice-i-cecile:
This is as good as we're going to get for an initial impl: merging and shipping.

PR #313, opened 2023-02-04
Fix `uninlined_format_args` lint

TimJentzsch:
Fixes the `uninlined_format_args` lint to make CI pass again.


alice-i-cecile (approved):
Many thanks.

PR #314, opened 2023-02-06
add many to one and one to many insert methods

PhaestusFox:
added two helper methods to make it simpler to add many-to-one mappings


alice-i-cecile (changes requested):
I'm happy with this :) Can you add a note to the Changelog under Usability?

alice-i-cecile (approved):
Merging, we'll add it to the changelog later.

PR #315, opened 2023-02-06
Fix incorrect directions

rlopzc:
Left should be west and right should be east


alice-i-cecile:
Thanks for the fix :)

PR #327, opened 2023-03-10
Add common conditions

Shatur:
Mirrors https://github.com/bevyengine/bevy/blob/main/crates/bevy_input/src/common_conditions.rs

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -6,7 +6,6 @@
 
 - Added common run conditions for actions that mirrors input conditions in Bevy.
 
-
 ## Version 0.9.0
 
 ### Usability
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -4,7 +4,8 @@
 
 ### Usability
 
-- Added common conditions for actions that mirrors input conditions in Bevy.
+- Added common run conditions for actions that mirrors input conditions in Bevy.
+
 
 ## Version 0.9.0
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,11 @@
 # Release Notes
 
+## Unreleased
+
+### Usability
+
+- Added common conditions for actions that mirrors input conditions in Bevy.
+
 ## Version 0.9.0
 
 ### Usability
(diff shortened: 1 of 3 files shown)


src/common_conditions.rs
```diff
@@ -0,0 +1,40 @@
+//! Run conditions for actions.
+
+use crate::{prelude::ActionState, Actionlike};
+use bevy::prelude::Res;
+
+/// Stateful run condition that can be toggled via an action press using [`ActionState::just_pressed`].
+pub fn action_toggle_active<T>(default: bool, action: T) -> impl FnMut(Res<ActionState<T>>) -> bool
+where
+    T: Actionlike + Copy,
```
alice-i-cecile, line 9: I'd like to relax these bounds to Clone: there's no reason to enforce Copy here.

RELEASES.md
```diff
@@ -1,5 +1,11 @@
 # Release Notes
 
+## Unreleased
+
+### Usability
+
+- Added common conditions for actions that mirrors input conditions in Bevy.
```
alice-i-cecile, line 7:
```suggestion
- Added common run conditions for actions that mirrors input conditions in Bevy.

```

src/common_conditions.rs
Shatur: But we will clone it every usage because `pressed` and similar take by value. Is it okay?

alice-i-cecile: Not any worse than the `Copy` bound :) In either case, it's getting copied every frame.

Shatur:
Looks like CI fails, but the error is unrelated to the PR.

alice-i-cecile (approved):

Shatur:
Right, I thought that markdown lint failed because of this:
```
 2023-03-20 23:16:49 [ERROR]   Failed to get [GITHUB_TOKEN]!
2023-03-20 23:16:49 [ERROR]   []
2023-03-20 23:16:49 [ERROR]   Please set a [GITHUB_TOKEN] from the main workflow environment to take advantage of multiple status reports!
```
But the actual error was this:
```
2023-03-10 21:40:29 [ERROR]   Found errors in [markdownlint] linter!
2023-03-10 21:40:29 [ERROR]   Error code: 1. Command output:
------
/github/workspace/RELEASES.md:9 MD012/no-multiple-blanks Multiple consecutive blank lines [Expected: 1; Actual: 2]
------
```

Issue #321, opened 2023-03-02
Represent user inputs via traits, rather than a large enum


alice-i-cecile:
The exhaustive approach is lovely, but a) leads to some very tricky nesting b) isn't user extensible c) results in excessive memory use, as every input requires the maximum amount of space.

In addition:

>               I think this will be obsolete if we use traits instead of enums to represent the user input, which is probably the better move in the long run.

_Originally posted by @TimJentzsch in https://github.com/Leafwing-Studios/leafwing-input-manager/issues/241#issuecomment-1451789336_

TimJentzsch:
This could also open the way for some changes to the user-facing API, for example we could implement the trait for tuples of user inputs to represent the chords, e.g. `(KeyCode::Shift, KeyCode::Alt, KeyCode::M)`

paul-hansen:
So I ran into pretty quickly that we need InputLike to be able to be serialized but adding a Serialize bound to the trait prevents using dynamic dispatch, which we need in order to store them in the same vec. E.g. `Vec<Box<InputLike>>` would return the error "cannot be made into an object".

Usually when I run into this I just switch to using enums like we already are, but that doesn't allow users of the library extend it. So keep in mind if we can find another way to support user extendibility or ditch that requirement, we could stick with an enum and probably do things like box the Chord variant to reduce the size of UserInput.

For solving the need for InputLike to be cloneable and comparable, I was able to add a clone_dyn and eq_dyn function to the trait so it could be implemented manually for the trait ([code here](https://github.com/paul-hansen/leafwing-input-manager/blob/dc00333c792d19605d85a5147e5f3d6170c269a0/src/user_input.rs#L79-L85)), but that doesn't work for Serialize because the serialize function needs to take a generic Serializer so adding a `serialize<S>(&self, serializer: S)` function to the InputLike trait would prevent using dynamic dispatch again.

I have a branch with my current attempt where I'm running into this issue here: https://github.com/Leafwing-Studios/leafwing-input-manager/compare/main...paul-hansen:leafwing-input-manager:userinput-as-trait There are probably major crimes committed in the code but I'm currently just trying to get something to compile as a proof of concept. This code currently errors with: 
```
error[E0277]: the trait bound `dyn InputLike: action_state::_::_serde::Serialize` is not satisfied
    --> src\input_map.rs:531:13
     |
529  |           input_map.serialize_field(
     |                     --------------- required by a bound introduced by this call
530  |               "map",
[...]

PR #331, opened 2023-03-26
Add Reflect and FromReflect to UserInput

hammypants:
Related to the contributing guidelines: I note that the `dev` branch is _very_ stale and I couldn't find any recent PRs against it, so I made this against `main`. I can adjust this if necessary.

## General

This PR adds `Reflect` and `FromReflect` to `UserInput` via adding it to the entire chain of types under its variants that also needed them. It was mentioned in a few places that this was desired, notably in https://github.com/Leafwing-Studios/leafwing-input-manager/pull/308, and I also would like to see this. (It kinda feels like any type that is intended to be a bevy component should have them by default and _not having_ them be opt-out, but I digress.)

## PetitSet ...

The big hang up here was that `PetitSet` does not implement either of said traits. I naively newtype-d it and delegated a majority of its features, but I would imagine there could be a need to have more careful thought put into this. It currently is just a poor veneer over `PetitSet`

`PetitSet` is currently exposed as API surface in a couple of places; so, this is kind of a big change. I can imagine a world where its not, though, and that the functionality of an input chord is much more pared down, or perhaps looks entirely different than a front for `PetitSet`, and this would perhaps force that thought into the light earlier than expected.

A perhaps less obtrusive approach would be to only do this behind the scenes and maintain `PetitSet` as the forward API. You all own both crates so I defer all decisions!

Last thing on the newtype: it's added to a separate module, but reexported through `user_input`-- this was entirely a "idk lmao" decision.

### DIFF
--- a/src/axislike.rs
+++ b/src/axislike.rs
@@ -20,7 +20,7 @@ use serde::{Deserialize, Serialize};
 /// # Warning
 ///
 /// `positive_low` must be greater than or equal to `negative_low` for this type to be validly constructed.
-#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, Reflect, FromReflect)]
 pub struct SingleAxis {
     /// The axis that is being checked.
     pub axis_type: AxisType,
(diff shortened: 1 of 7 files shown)


alice-i-cecile:
Definitely in favor of having these traits implemented! I'm actually planning on dropping petitset as a dependency, in favor of a simple hash or btree map in #321.

As a result, I'm a bit reluctant to merge this in the current form, but I'm open to suggestions on the best way to get to where we're hoping to be.

alice-i-cecile:
Good analysis: I was thinking of using one of the blessed types here :)

If you're interested in moving this forward faster, I'd be happy to merge a PR migrating to hash maps, and then we can rebase and merge this?

hammypants:
> Good analysis: I was thinking of using one of the blessed types here :)
> 
> If you're interested in moving this forward faster, I'd be happy to merge a PR migrating to hash maps, and then we can rebase and merge this?

I can do that!

alice-i-cecile:
Fixing as part of #400.

hammypants:
> Fixing as part of #400.

My apologies, I was hoping https://github.com/Leafwing-Studios/leafwing-input-manager/issues/321 would have landed as [at the time] removing `PetitSet` was looking more involved than I had time for and this seemed close.

PR #338, opened 2023-04-07
Replace unmaintained actions-rs/toolchain action in CI workflows

striezel:
Basically all of the `actions-rs/*` actions are unmaintained. See <https://github.com/actions-rs/toolchain/issues/216> for more information. Due to their age they generate several warnings in CI runs, for example in https://github.com/Leafwing-Studios/leafwing-input-manager/actions/runs/4537080312:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions-rs/toolchain@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

To get rid of those warnings the occurrences of `actions-rs/toolchain` are replaced by [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,11 +14,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
           components: rustfmt, clippy
-          override: true
       - name: Cache Cargo build files
         uses: Leafwing-Studios/cargo-cache@v1.0.0
       - name: Install alsa and udev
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -14,11 +14,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
```
TimJentzsch, line 17: Perhaps we should use `@stable` here instead of `@master`.
At least that's [what Bevy is currently doing](https://github.com/bevyengine/bevy/blob/3ead10a3e0ab1c23d8e4b6cc9fa8be1eef8c581f/.github/workflows/ci.yml#L32)

striezel: Using `dtolnay/rust-toolchain@stable` is just a shortcut for
```
uses: dtolnay/rust-toolchain@master
with:
    toolchain: stable
```
which is what the current changes in this PR do. `dtolnay/rust-toolchain` has the feature that you can basically specify the Rust version via the version of the action, if it is not specified via the `toolchain` parameter. Or in other words: Using `dtolnay/rust-toolchain@stable` is just another way of saying _"Please install and use the current stable Rust toolchain."_, and that is what both variants do here.

TimJentzsch (approved):

PR #341, opened 2023-04-10
Fix various errors in deadzone calculation/filtering

ndarilek:
I noticed that a number of deadzones weren't correctly being calculated. In particular, inputs with values like (1, 0.05) and a deadzone of 0.1 would return (1, 0.05) for the axis values when I believe they should have returned (1, 0). I added initially failing tests that demonstrated this behavior.

Additionally, calculating whether a dual-axis input was pressed or released simply looked at its value;. It didn't check whether each member axis was above the threshold.

Finally, there's some confusion in my mind about `positive_low` and `negative_low`. In particular I think `negative_low` should always be negative, tested by assertion. `SingleAxis::negative_only` was just passing the same deadzone in for both values and failing the additional calculations I'd added. I struggled a bit with this until I discovered I was passing a positive value in for the threshold. Fixing that got my controls working as I'd expect, but ideally the API should prevent this.

I think this PR is an improvement, but it may not correctly represent the intent behind `negative_low`. If it should always be positive and negated for testing, I'd like to see assertions somewhere so the intent is clear. Happy to add those here or elsewhere, but since structs can be created from anywhere, I'm not immediately sure what the best place for that check would be.

Thanks.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -100,7 +100,7 @@
 - Added gamepad axis support.
   - Use the new `SingleAxis` and `DualAxis` types / variants.
 - Added mousewheel and mouse motion support.
-  - Use the new `SingleAxis` and `DualAxis` types / variants when you care about the continous values.
+  - Use the new `SingleAxis` and `DualAxis` types / variants when you care about the continuous values.
   - Use the new `MouseWheelDirection` enum as an `InputKind`.
 - Added `SingleAxis` and `DualAxis` structs that can be supplied to an `InputMap` to trigger on axis inputs.
 - Added `VirtualDPad` struct that can be supplied to an `InputMap` to trigger on four direction-representing inputs.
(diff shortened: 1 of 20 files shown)


ndarilek:
Hey, just wanted to check on this one.

Also, contributing.md mentions development happening on the dev branch, but it looks like that hasn't been updated since last year? Just want to make sure I'm not missing something.

I've been testing this branch and it seems fine, but it's possible I've missed something or made a mistake. Hoping to cut a release of my game with it on Monday or Tuesday.

Thanks.

alice-i-cecile:
That's correct: that practice has been abandoned. Let me review this PR now; I've been a little burnt out <3

alice-i-cecile (approved):
This LGTM: those are both bugs. Can you update CHANGELOG.md briefly to reflect your changes?

ndarilek:
Sure thing. Sorry, been there re: burnout and didn't mean to rush you.

ndarilek:
Done. There wasn't a top-level section for the new release so I just added the update (hopefully) correctly-structured and near where it should go. Hope that's sufficient.

Thank you!

RELEASES.md
```diff
@@ -1,5 +1,10 @@
 # Release Notes
 
+### Bugs
```
alice-i-cecile, line 3:
```suggestion

## Unreleased

### Bugs

```

alice-i-cecile:
Just a minor formatting nit, but I've fixed that up for you. Merging!

PR #346, opened 2023-05-16
UserInput as trait

paul-hansen:
### Description
Change our UserInput enum to be a trait. Adds the ability for other libraries to add new inputs that can be mapped to our actions. Also removes all UserInput instances being the size of the largest variant (which was inflated due to Chords using PetiteSets stored on the stack)

Closes: #321

### Current Status
Good progress has been made but still some major things to finish implementing such as deserialization.

### TODO
- [x] Implement the new InputLike and relevant ButtonLike/AxisLike traits for the inputs we previously supported
    For reference: [old InputStreams](https://docs.rs/leafwing-input-manager/0.9.2/leafwing_input_manager/input_streams/struct.InputStreams.html) &  [old InputKind](https://docs.rs/leafwing-input-manager/0.9.2/leafwing_input_manager/user_input/enum.InputKind.html)
   - [x] Gamepads
   - [x] KeyCode
   - [x] ScanCode
   - [x] MouseButton
   - [x] MouseWheel
   - [x] MouseMotion
   - [x] VirtualDPad & VirtualAxis
   - [x] Chords
- [x] Remove Input Mocking ([see comment](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/321#issuecomment-1548827763))
- [x] Update branch to Bevy 0.11
- [ ] Figure out how to handle deadzones (e.g. is it handled by the implementor of inputlike, purely in our lib, or some mix of the two?)
- [ ] Reimplement egui and ui features[ ignored input](https://github.com/Leafwing-Studios/leafwing-input-manager/blob/887af2fec9838e3ce907c06b6fbd01fb9b972985/src/systems.rs#L96-L108)
- [ ] Address all added TODO comments
- [ ] Fix tests
- [ ] Review project/folder structure (e.g. I currently have a different module for each InputLike implementation, maybe we don't want that)
- [ ] Documentation and re-enable `#![forbid(missing_docs)]`
- [ ] Performance testing / Comparison / Benchmarks
- [ ] Test all examples
- [ ] Migration Guide


src/input_map.rs
```diff
@@ -78,7 +78,7 @@ use std::marker::PhantomData;
 pub struct InputMap<A: Actionlike> {
     /// The raw vector of [PetitSet]s used to store the input mapping,
     /// indexed by the `Actionlike::id` of `A`
-    map: Vec<PetitSet<UserInput, 16>>,
+    map: Vec<PetitSet<Box<dyn InputLikeObject>, 16>>,
```
alice-i-cecile, line 81: We should just swap to a HashMap here now that we're no longer stack-allocating. Will simplify a *lot* of code.

paul-hansen:
Rebased on main to fix conflict.

The `minimal` and `physical_key_bindings` examples are working again!

I think something that needs some thought is ButtonLike and AxisLike, currently they are just marker traits that could easily be replaced with a bool. I feel like they could be doing more. Maybe split up the InputStreams Trait into those traits or something, not sure if that's possible yet. I'll probably implement Gamepad next since that has axes, and might give some insight.

src/input_map.rs
paul-hansen: I assume we still want "slots" so the inputs are ordered and we can have a "secondary" input in index 1 without the "primary" index 0 being populated. Probably accomplish this with a `HashMap<usize, Box<dyn InputLikeObject>`?

paul-hansen: What would you think of having a type that wraps the HashMap so we can make it more clear what the purpose of the usize is? Named something `Bindings` maybe? That way when we call things like [InputMap::get](https://docs.rs/leafwing-input-manager/latest/leafwing_input_manager/input_map/struct.InputMap.html#method.get) we return our Bindings struct instead of a hashmap with a usize as key that is somewhat unclear what the purpose is.

alice-i-cecile:
I would expect we want a method to create a `Buttonlike` input from an `Axislike` input :) We may also want to split it into `Buttonlike` / `SingleAxislike` / `DualAxislike` instead: I think that the methods I really want from something that behaves like an axis tend to depend on the number of dimensions.

paul-hansen:
@alice-i-cecile could use some input on the direction you want to go with deadzones. It used to be that [SingleAxis](https://docs.rs/leafwing-input-manager/latest/leafwing_input_manager/axislike/struct.SingleAxis.html) would store the positive_low and negative_low values for deadzone. Now that SingleAxisLike is a trait, we can implement it for [GamepadAxis](https://docs.rs/bevy/latest/bevy/input/gamepad/struct.GamepadAxis.html) and directly insert that into the mapping instead of converting it into a SingleAxis first which seems like a win. But now it's unclear where to handle the deadzone. 

Here's the solutions I've thought of so far:
1. Let GamepadAxis be inserted directly and don't do anything with deadzones in that case, but keep SingleAxis around as a wrapper that can optionally be used to handle deadzones.
2. Try to prevent GamepadAxis being inserted directly into the input map and force the user to wrap it in SingleAxis first so they have to set the deadzones, e.g. making SingleAxisLike not implement InputLikeObject 
3. Add methods to the SingleAxisLike trait for setting the deadzone values and let the implementation handle what it does with those values. For example, for GamepadAxis it would update Bevy's AxisSettings

Feel free to take some time to respond, I have other things I can keep working on with this.

paul-hansen:
Initial benchmarks look pretty good. Doesn't necessarily mean anything since not all the tests are passing yet.
<details>
  <summary>Click to expand benchmarks</summary>

```
Running benches\action_state.rs (target\release\deps\action_state-c2b1817d6cb7d760.exe)
Gnuplot not found, using plotters backend
action_state_default    time:   [125.01 ns 127.13 ns 129.81 ns]
                        change: [+7.8423% +10.838% +14.549%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 16 outliers among 100 measurements (16.00%)
  16 (16.00%) high severe

pressed                 time:   [189.91 ps 190.63 ps 191.30 ps]
                        change: [+0.9657% +1.3970% +1.8094%] (p = 0.00 < 0.05)
                        Change within noise threshold.

just_pressed            time:   [187.10 ps 187.61 ps 188.24 ps]
                        change: [+1.1177% +1.4725% +1.8642%] (p = 0.00 < 0.05)
                        Performance has regressed.

released                time:   [186.08 ps 186.34 ps 186.62 ps]
                        change: [+0.2191% +0.4256% +0.6172%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe

just_released           time:   [191.08 ps 191.52 ps 192.00 ps]
                        change: [+3.1319% +3.4464% +3.7604%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  2 (2.00%) low severe
  3 (3.00%) high mild
  4 (4.00%) high severe

update                  time:   [96.070 ns 96.316 ns 96.618 ns]
                        change: [-1.8802% -0.7222% +0.2646%] (p = 0.20 > 0.05)
                        No change in performance detected.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

     Running benches\input_map.rs (target\release\deps\input_map-9caee25e56b382f0.exe)
Gnuplot not found, using plotters backend
Benchmarking construct_input_map_from_iter: Collecting 100 samples in estimated 5.0003 s (8.3M iconstruct_input_map_from_iter
                        time:   [604.00 ns 606.87 ns 609.46 ns]
                        change: [-65.687% -65.516% -65.351%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking construct_input_map_from_chained_calls: Collecting 100 samples in estimated 5.0030 construct_input_map_from_chained_calls
                        time:   [1.0426 µs 1.0441 µs 1.0461 µs]
                        change: [-85.377% -85.194% -85.005%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) low mild
  2 (2.00%) high severe

Benchmarking which_pressed/PressAll: Collecting 100 samples in estimated 5.0100 s (1.7M iteratiowhich_pressed/PressAll  time:   [2.9919 µs 2.9946 µs 2.9976 µs]
                        change: [-36.098% -35.773% -35.474%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  1 (1.00%) low severe
  6 (6.00%) high mild
  4 (4.00%) high severe
Benchmarking which_pressed/PrioritizeLongest: Collecting 100 samples in estimated 5.0000 s (1.7Mwhich_pressed/PrioritizeLongest
                        time:   [2.9454 µs 2.9520 µs 2.9600 µs]
                        change: [-43.600% -43.032% -42.523%] (p = 0.00 < 0.05)
  1 (1.00%) high mild
  1 (1.00%) high severe
Benchmarking which_pressed/UseActionOrder: Collecting 100 samples in estimated 5.0032 s (1.7M itwhich_pressed/UseActionOrder
                        time:   [2.9275 µs 2.9330 µs 2.9385 µs]
                        change: [-43.848% -43.526% -43.237%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high severe

```

</details>

alice-i-cecile:
Preference is for 3 if we can swing it: it definitely feels like managing deadzones is within the domain of the trait.

src/clashing_inputs.rs
```diff
@@ -452,55 +322,51 @@ mod tests {
...
             assert!(!ab.clashes(&bc));
             assert!(ab.clashes(&abc));
-            assert!(axyz_dpad.clashes(&a));
+            // VirtualDPads are considered single inputs so they don't clash with other single inputs
+            assert!(!axyz_dpad.clashes(&A));
             assert!(axyz_dpad.clashes(&ab));
             assert!(!axyz_dpad.clashes(&bc));
-            assert!(axyz_dpad.clashes(&abcd_dpad));
+            assert!(!axyz_dpad.clashes(&abcd_dpad));
             assert!(ctrl_up.clashes(&directions_dpad));
```
paul-hansen, line 370: @alice-i-cecile I inverted two assertions here. It seemed to make more sense to me and was easier to implement if VirtualDPads don't clash with other single inputs since they are considered having a length of 1 and other single inputs don't clash with each other.

It seems like the concept of "clashes" is more accurately "conflicts" where it's purpose isn't to prevent the same input from being mapped to other things, but to detect when a key combination overlaps another input and prioritize longer combinations. For virtual dpads it seems like it would be more consistent to treat them the same as single inputs in regards to clashes. 

Let me know if there was a reason for this or if you don't agree and I can work in an exception to how they are handled.

alice-i-cecile: Yeah, that seems like a better approach.

paul-hansen:
I might not get back to this seriously before mid September, working more hours to get a demo done for our game for a convention. If anyone wants to pick it up before then I'd be happy to answer any questions.

Current status is that it passes all the tests except for the deadzone ones because that isn't implemented yet. Checklist on original post has been updated with some other things that need finished like docs, migration guide etc.
Deadzones might require some bigger changes, especially if we want to keep the functionality from https://github.com/Leafwing-Studios/leafwing-input-manager/pull/370 since I think we would have to store something alongside inputs for the deadzone settings instead of only storing the trait object we get from integrations.

paul-hansen:
I think I'll have to bow out of heading this up. I've been trying to merge or rebase from main but there are enough conflicting changes it might be cleaner to start over and just look at the [`InputLike`](https://github.com/Leafwing-Studios/leafwing-input-manager/pull/346/files#diff-27ef912b17b0ea897c219c4c22427afd1ad9221301469df14554bfc5e89175b4R24) and [`InputLikeObject`](https://github.com/Leafwing-Studios/leafwing-input-manager/pull/346/files#diff-27ef912b17b0ea897c219c4c22427afd1ad9221301469df14554bfc5e89175b4R30) traits from this PR as a starting point. As it stands, I don't think I can spare the time to commit to that sadly. In hindsight, it might be worth considering a feature freeze or forking the project as a new crate or something while taking on this much of a refactor to avoid having to re-implement changes that gets added while working on it. 

Hopefully this was still valuable as an experiment to show a way it could be done using traits.

Closing this PR for now, but if anyone wants to try and resurrect the branch as their own or anything you're more than welcome to use it. I'm also open to doing work related to it if there are ideas on how to make it simpler or break it into smaller changes that could be merged incrementally.

alice-i-cecile:
Sounds good <3 thanks for doing this investigation; I'll take a crack at this next time I'm in the mood for something crunchy.

PR #360, opened 2023-06-27
Mouse Position Example and Multiple Targets for Action State Driver

jamescarterbell:
# Problem #

A lot of games require mouse position as an input in some way.  However, in games with multiple windows or screens, this is not a straight forward input to get, and how to transform from screen space to world space, or if you need to at all, is a bit dependent on the use case, so it's hard to have a one size fits all axis type.

# Solution #

This crates great design makes it relatively easy to drive a mouse position action from an ActionStateDriver, it just needs a quick example to get the ball rolling for how you can add the capability to your own game, which this PR adds.

This PR also makes a small change to ActionStateDriver to let it drive multiple entities instead of just one.  I think this will be a big win long term for games that will need mouse data, or other driven action data, on many entities.


alice-i-cecile:
Change log please :) Also the promised example appears to be missing?

PR #361, opened 2023-06-27
Mouse position

jamescarterbell:
Problem
A lot of games require mouse position as an input in some way. However, in games with multiple windows or screens, this is not a straight forward input to get, and how to transform from screen space to world space, or if you need to at all, is a bit dependent on the use case, so it's hard to have a one size fits all axis type.

Solution
This crates great design makes it relatively easy to drive a mouse position action from an ActionStateDriver, it just needs a quick example to get the ball rolling for how you can add the capability to your own game, which this PR adds.

This PR also makes a small change to ActionStateDriver to let it drive multiple entities instead of just one. I think this will be a big win long term for games that will need mouse data, or other driven action data, on many entities.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -9,7 +9,11 @@ to update multiple entities if needed.
 
 ### Docs
 
-- Added example for getting driving a cursor position action.
+- Added example for driving cursor position action from another entity.
+
+### Usability
+
+- Makes `run_if_enabled` public.
 
 ## Version 0.9.3
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bugs
+
+- Changed `Rotation` to be stored in millionths of a degree instead of tenths of a degree in order to reduce rounding errors.
+
 ### Usability
 
 - Added `VirtualAxis::horizontal_dpad()` and `VirtualAxis::vertical_dpad()`.
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,16 @@
 # Release Notes
 
+## Unreleased
+
+### Enhancements
+
+- Changed `entity` field of `ActionStateDriver` to `targets: ActionStateDriverTarget` with variants for 0, 1, or multiple targets, to allow for one driver 
+to update multiple entities if needed.
+
+### Docs
+
+- Added example for getting driving a cursor position action.
+
 ## Version 0.9.3
 
 ### Bugs
--- a/examples/mouse_position.rs
+++ b/examples/mouse_position.rs
@@ -9,7 +9,7 @@ fn main() {
         .add_plugin(InputManagerPlugin::<BoxMovement>::default())
         .add_startup_system(setup)
         .add_system(
-            update_cursor_state_from_window::<BoxMovement>
+            update_cursor_state_from_window
                 .run_if(run_if_enabled::<BoxMovement>)
                 .in_base_set(CoreSet::PreUpdate)
                 .in_set(InputManagerSystem::ManualControl)
@@ -24,7 +24,7 @@ fn main() {
 
 #[derive(Actionlike, Clone, Debug, Copy, PartialEq, Eq)]
 enum BoxMovement {
-    Pan,
+    MousePosition,
 }
 
 fn setup(mut commands: Commands, window: Query<Entity, With<PrimaryWindow>>) {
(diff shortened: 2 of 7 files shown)


RELEASES.md
```diff
@@ -1,5 +1,16 @@
...
+## Unreleased
+
+### Enhancements
+
+- Changed `entity` field of `ActionStateDriver` to `targets: ActionStateDriverTarget` with variants for 0, 1, or multiple targets, to allow for one driver 
+to update multiple entities if needed.
+
+### Docs
+
+- Added example for getting driving a cursor position action.
```
alice-i-cecile, line 12: "getting driving"?

examples/mouse_position.rs
```diff
@@ -0,0 +1,84 @@
...
+                .after(InputManagerSystem::Tick)
+                .after(InputManagerSystem::Update)
+                .after(InputSystem),
+        )
+        .add_system(pan_camera)
+        .run();
+}
+
+#[derive(Actionlike, Clone, Debug, Copy, PartialEq, Eq)]
+enum BoxMovement {
```
alice-i-cecile, line 26: A clearer name of maybe some doc strings would help users understand what this is being used for faster.

examples/mouse_position.rs
```diff
@@ -0,0 +1,84 @@
...
+        // Note: another entity will be driving this input
+        .id();
+
+    commands.entity(window.single()).insert(ActionStateDriver {
+        action: BoxMovement::Pan,
+        targets: entity.into(),
+    });
+}
+
+fn update_cursor_state_from_window<A: Actionlike>(
```
alice-i-cecile, line 48: Hmm, it's nice to demonstrate the power of generics but a bit weird when we're only using a single type in the example.

Maybe a comment here calling out what `A` is?

examples/mouse_position.rs
```diff
@@ -0,0 +1,84 @@
...
+    window_query: Query<(&Window, &ActionStateDriver<A>)>,
+    mut action_state_query: Query<&mut ActionState<A>>,
+) {
+    // Update each actionstate with the mouse position from the window
+    for (window, driver) in window_query.iter() {
+        for entity in driver.targets.iter() {
```
alice-i-cecile, line 54: A reminder about how `ActionStateDriver` works would be helpful here.

src/action_state.rs
```diff
@@ -537,20 +539,140 @@ impl<A: Actionlike> Default for ActionState<A> {
...
+    /// Single target
+    Single(Entity),
+    /// Multiple targets
+    Multi(HashSet<Entity>),
+}
+
+impl ActionStateDriverTarget {
+    /// Get an iterator for the entities targeted.
+    #[inline(always)]
+    pub fn iter(&self) -> impl Iterator<Item = &Entity> {
```
alice-i-cecile, line 572: Returning a `&Entity` as the item is weird here because it's a copy type.

src/action_state.rs
```diff
@@ -537,20 +539,140 @@ impl<A: Actionlike> Default for ActionState<A> {
...
+                ActionStateDriverTargetIterator::Single(std::iter::once(entity))
+            }
+            Self::Multi(entities) => ActionStateDriverTargetIterator::Multi(entities.iter()),
+        }
+    }
+
+    /// Insert an entity as a target.
+    #[inline(always)]
+    pub fn insert(&mut self, entity: Entity) {
+        *self = std::mem::replace(self, Self::None).with(entity);
```
alice-i-cecile, line 585: The mem::replace is tricky to understand (and verify) here: why use it?

src/action_state.rs
```diff
@@ -537,20 +539,140 @@ impl<A: Actionlike> Default for ActionState<A> {
...
+    }
+
+    /// Remove an entity as a target if it's in the set using a builder style pattern.
+    pub fn without(mut self, entity: Entity) -> Self {
+        match self {
+            Self::None => Self::None,
+            Self::Single(_) => Self::None,
+            Self::Multi(ref mut targets) => {
+                targets.remove(&entity);
+                match targets.len() {
```
alice-i-cecile, line 638: This is probably nicer if you just impl `FromIter` and then use that trait here.

src/systems.rs
```diff
@@ -287,6 +289,6 @@ pub fn release_on_input_map_removed<A: Actionlike>(
 }
 
 /// Uses the value of [`ToggleActions<A>`] to determine if input manager systems of type `A` should run.
-pub(super) fn run_if_enabled<A: Actionlike>(toggle_actions: Res<ToggleActions<A>>) -> bool {
+pub fn run_if_enabled<A: Actionlike>(toggle_actions: Res<ToggleActions<A>>) -> bool {
```
alice-i-cecile, line 292: This should be in the changelog.

alice-i-cecile (changes requested):
Code quality / explanation nits, but I'm on board with the direction!

examples/mouse_position.rs
jamescarterbell: I just copied and pasted the ui signature, going to just remove the generic here.

src/action_state.rs
jamescarterbell: Keeping due to conversation about holding reference to underlying collection.

src/action_state.rs
jamescarterbell: I don't want to copy the with code and make minor changes for working with a reference, but replacing a mutable ref in place is a bit of a pain

alice-i-cecile: Ah, fair enough. Just drop a comment in the source then please.

jamescarterbell:
I think I made all the changes requested.

alice-i-cecile (approved):

PR #362, opened 2023-06-30
Add Inversion to SingleAxis, DualAxis, and VirtualDPad

### LINKED ISSUE #247: Adding inversion to mouse and gamepad axes

I want to implement keyboard and mouse control for a small game. I would like to give the user the choice of inverting the mouse axes independently from the keyboard input. In ego shooters some people like it to move the mouse forward to look up, others move the mouse backwards to look up.

When setting the `InputMap` it would be nice to be able to set some axes to inverted.

A boolean `inverted` field in `SingleAxis` might already provide this feature. This can be set to `true` to mark this axis as being inverted.

## What alternatives have you considered?

Currently I implement custom `PitchInverted` action to represent the inverted axis for mouse only. This is a workaround and I would prefer a built-in solution, that is easily switchable by user choice.

## Related work

Using `SingleAxis` to solve this would enable inversion not only for mouse but also for all gamepad axes. Inversion could be useful for them as well.

SignalWalker:
Fixes #247.

Adds `inverted`, `inverted_x`, and `inverted_y` builder-style functions for `SingleAxis`, `DualAxis`, and `VirtualDPad`, where appropriate. The `SingleAxis` implementation uses an `inverted` field, the `DualAxis` implementation just inverts its internal axes, and the `VirtualDPad` implementation swaps the relevant inputs.

Current tests pass, but I haven't written any yet to test inversion.

## Todo
- [ ] Write tests
- [ ] Better documentation?

### DIFF
--- a/src/axislike.rs
+++ b/src/axislike.rs
@@ -28,6 +28,8 @@ pub struct SingleAxis {
     pub positive_low: f32,
     /// Any axis value lower than this will trigger the input.
     pub negative_low: f32,
+    /// Whether to invert output values from this axis.
+    pub inverted: bool,
     /// The target value for this input, used for input mocking.
     ///
     /// WARNING: this field is ignored for the sake of [`Eq`] and [`Hash`](std::hash::Hash)
@@ -42,6 +44,7 @@ impl SingleAxis {
             axis_type: axis_type.into(),
             positive_low: threshold,
             negative_low: -threshold,
+            inverted: false,
             value: None,
         }
     }
@@ -56,6 +59,7 @@ impl SingleAxis {
             axis_type: axis_type.into(),
             positive_low: 0.0,
             negative_low: 0.0,
+            inverted: false,
             value: Some(value),
         }
     }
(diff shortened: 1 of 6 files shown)


src/axislike.rs
```diff
@@ -250,6 +267,28 @@ impl DualAxis {
...
+    }
+
+    /// Returns this [`DualAxis`] with an inverted Y-axis.
+    #[must_use]
+    pub fn inverted_y(mut self) -> DualAxis {
+        self.y = self.y.inverted();
+        self
+    }
+
+    /// Returns this [`DualAxis`] with inverted axes.
```
alice-i-cecile, line 285:
```suggestion
    /// Returns this [`DualAxis`] with both axes inverted.
```

alice-i-cecile (changes requested):
Solid :) Changelog entry and a couple tests and this looks good to me.

SignalWalker:
> Solid :) Changelog entry and a couple tests and this looks good to me.

Cool, thanks. `CONTRIBUTING.md` says to put changes into `CHANGELOG.md`, which doesn't exist -- I assume it should be `RELEASES.md`?

alice-i-cecile:
That's correct. Thanks for pointing that out; I'll update it!

alice-i-cecile (approved):

PR #366, opened 2023-07-13
Remove Actionlike TypePath trait requirement

paul-hansen:
There's a [PR open with Bevy](https://github.com/bevyengine/bevy/pull/9140) that allows opting out of the derrived TypePath implementation from Reflect which allows us to implement it manually, this allows us to remove the TypePath restriction we had to add for the Bevy 0.11 update.

## TODO
- [ ] Remove patch once https://github.com/bevyengine/bevy/pull/9140 is merged and released. 
- [ ] Change bevy dependency to require the patch that includes 9140 so people that pin the 0.11.0 Bevy release must update Bevy first.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,3 +67,5 @@ harness = false
 name = "leafwing_input_manager"
 path = "src/lib.rs"
 
+[patch.crates-io]
+bevy = { git = "https://github.com/paul-hansen/bevy.git", rev = "d7ba279f"}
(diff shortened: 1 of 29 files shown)


paul-hansen:
I'm actually leaning against this now personally, though I'm open to discussion. 

The reason Bevy switched to using the TypePath trait instead of relying on Rust's [type_name](https://doc.rust-lang.org/std/any/fn.type_name.html) is that type_name is provides no guarantee of it's output staying the same between different versions of the compiler. This means that if we used type_name and we serialized the input map to a file, that file could break going from one version of Rust to the next, without any lines of code in the project being changed. Since serializing the input map to a file is a primary goal of LWIM I think that's pretty important for us for the same reasons it's important for Bevy.

alice-i-cecile:
Good analysis, I agree with that assessment.

PR #368, opened 2023-07-24
Readme example

rparrett:
These got out of sync with the Bevy 0.11 update, and the current code in README doesn't compile.


TimJentzsch (approved):

TimJentzsch:
We'll have to merge #369 before this one to make CI pass

PR #370, opened 2023-07-30
Deadzone shapes

100-TomatoJuice:
Currently, the only type of deadzone for a `DualAxis` is a cross shape. This shape is preferred for some instances, like in Rocket League, as the cross-shape deadzone makes directional flipping easier. However, there are instances where a circular or rectangular deadzone is preferred, like a twin-stick shooter.

Adds `DeadzoneShape` to `DualAxis`, which allows the user to choose between a circle, rectangle, and a cross for the deadzone shape.

From manual tests, the shapes appear to be correct and all work. This pr also passed `cargo test`.

### Things to do:
- [x] Add tests to make sure the deadzones shapes are correct
- [x] Implement `Hash` for `DeadZoneShape`
- [x] Change tests that relied on `Cross` snapping
- [x] Improve documentation of `DeadzoneShape` (feels a little convoluted)
- [x] Update `RELEASES.md`

### Questions
1. Is `extract_dual_axis_data()` the correct method to test the different shapes?
2. Adding `include_deadzone: bool` to `input_value()` feels a little gross, but I didn't find a better way to get the input_value of a `SingleAxis` without it running its deadzone. Is there a better, perhaps overlooked, way of doing this?
3. Currently, the shapes diamensions rely on the x and y thresholds in `SingleAxis`. This keeps inverting of the input easy, but might create confusion? `with_deadzone()` has been changed to hopefully make this easy, but there might be a better way.

### DIFF
--- a/src/axislike.rs
+++ b/src/axislike.rs
@@ -705,9 +705,14 @@ impl From<DualAxisData> for Vec2 {
 }
 
 /// The shape of the deadzone for a [`DualAxis`] input.
+///
+/// Input values that are on the line of the shape are counted as inside
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
 pub enum DeadZoneShape {
-    /// Deadzone with the shape of a cross. The cross is represented by two rectangles.
+    /// Deadzone with the shape of a cross.
+    ///
+    /// The cross is represented by two rectangles. When using [`DeadZoneShape::Cross`],
+    /// make sure rect_1 and rect_2 do not have the same values, otherwise the shape will be a rectangle
     Cross {
         /// The width of the first rectangle.
         rect_1_width: f32,
@@ -757,7 +762,9 @@ impl DeadZoneShape {
                 *rect_2_height,
             ),
             DeadZoneShape::Rect { width, height } => self.outside_rectangle(x, y, *width, *height),
-            DeadZoneShape::Ellipse { radius_x, radius_y } => self.outside_ellipse(x, y, *radius_x, *radius_y),
+            DeadZoneShape::Ellipse { radius_x, radius_y } => {
+                self.outside_ellipse(x, y, *radius_x, *radius_y)
+            }
         }
     }
 
--- a/src/axislike.rs
+++ b/src/axislike.rs
@@ -1,10 +1,9 @@
 //! Tools for working with directional axis-like user inputs (gamesticks, D-Pads and emulated equivalents)
 
 use crate::buttonlike::{MouseMotionDirection, MouseWheelDirection};
-use crate::input_streams::InputStreams;
 use crate::orientation::{Direction, Rotation};
 use crate::prelude::QwertyScanCode;
-use crate::user_input::{InputKind, UserInput};
+use crate::user_input::InputKind;
 use bevy::input::{
     gamepad::{GamepadAxisType, GamepadButtonType},
     keyboard::KeyCode,
@@ -188,7 +187,7 @@ pub struct DualAxis {
     /// The axis representing vertical movement.
     pub y: SingleAxis,
     /// The shape of the deadzone
-    pub deadzone_shape: DeadZoneShape,
+    pub deadzone: DeadZoneShape,
 }
 
 impl DualAxis {
@@ -201,8 +200,10 @@ impl DualAxis {
     ///
     /// This cannot be changed, but the struct can be easily manually constructed.
     pub const DEFAULT_DEADZONE_SHAPE: DeadZoneShape = DeadZoneShape::Cross {
-        horizontal_width: Self::DEFAULT_DEADZONE,
-        vertical_width: Self::DEFAULT_DEADZONE,
+        rect_1_width: Self::DEFAULT_DEADZONE,
+        rect_1_height: Self::DEFAULT_DEADZONE,
+        rect_2_width: Self::DEFAULT_DEADZONE,
+        rect_2_height: Self::DEFAULT_DEADZONE,
     };
 
     /// Creates a [`DualAxis`] with both `positive_low` and `negative_low` in both axes set to `threshold` with a `deadzone_shape`.
@@ -216,7 +217,7 @@ impl DualAxis {
         DualAxis {
             x: SingleAxis::symmetric(x_axis_type, threshold),
             y: SingleAxis::symmetric(y_axis_type, threshold),
-            deadzone_shape,
+            deadzone: deadzone_shape,
         }
     }
 
@@ -234,7 +235,7 @@ impl DualAxis {
         DualAxis {
             x: SingleAxis::from_value(x_axis_type, x_value),
             y: SingleAxis::from_value(y_axis_type, y_value),
-            deadzone_shape: Self::DEFAULT_DEADZONE_SHAPE,
+            deadzone: Self::DEFAULT_DEADZONE_SHAPE,
         }
     }
 
@@ -265,7 +266,7 @@ impl DualAxis {
         DualAxis {
             x: SingleAxis::mouse_wheel_x(),
             y: SingleAxis::mouse_wheel_y(),
-            deadzone_shape: Self::DEFAULT_DEADZONE_SHAPE,
+            deadzone: Self::DEFAULT_DEADZONE_SHAPE,
         }
     }
 
@@ -274,14 +275,14 @@ impl DualAxis {
         DualAxis {
             x: SingleAxis::mouse_motion_x(),
             y: SingleAxis::mouse_motion_y(),
-            deadzone_shape: Self::DEFAULT_DEADZONE_SHAPE,
+            deadzone: Self::DEFAULT_DEADZONE_SHAPE,
         }
     }
 
     /// Returns this [`DualAxis`] with the deadzone set to the specified values and shape
     #[must_use]
     pub fn with_deadzone(mut self, deadzone: DeadZoneShape) -> DualAxis {
-        self.deadzone_shape = deadzone;
+        self.deadzone = deadzone;
         self
     }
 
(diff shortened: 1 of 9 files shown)


src/axislike.rs
```diff
@@ -688,3 +703,22 @@ impl From<DualAxisData> for Vec2 {
         data.xy
     }
 }
+
+/// The shape of the deadzone for a [`DualAxis`] input.
+/// 
+/// Uses the x and y thresholds on [`DualAxis`] to set the shape diamensions
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub enum DeadzoneShape{
+    /// Deadzone with a cross
```
alice-i-cecile, line 712: These docs need some improvement: what does the deadzone actually *look* like?

src/axislike.rs
```diff
@@ -688,3 +703,22 @@ impl From<DualAxisData> for Vec2 {
         data.xy
     }
 }
+
+/// The shape of the deadzone for a [`DualAxis`] input.
+/// 
+/// Uses the x and y thresholds on [`DualAxis`] to set the shape diamensions
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub enum DeadzoneShape{
```
alice-i-cecile, line 711: IMO we should merge the deadzone properties into this enum. That should improve clarity, and enable more exotic deadzone shapes with more or less parameters in the future.

src/axislike.rs
```diff
@@ -688,3 +703,22 @@ impl From<DualAxisData> for Vec2 {
...
+    /// 
+    /// Uses the x threshold as the horizontal width and y threshold as the vertical width.
+    Cross,
+    /// Deadzone with a rectangle.
+    /// 
+    /// Uses the x threshold as the width and y threshold as the length.
+    Rect,
+    /// Deadzone with a circle.
+    /// 
+    /// Uses the larger of the x or y threshold as the radius
```
alice-i-cecile, line 722: I think we should also have an ellipse. I could see an argument either way for replacing the circle completely, so I'll leave it up to you.

src/input_streams.rs
```diff
@@ -375,28 +375,49 @@ impl<'a> InputStreams<'a> {
...
-        if x > dual_axis.x.positive_low
-            || x < dual_axis.x.negative_low
-            || y > dual_axis.y.positive_low
-            || y < dual_axis.y.negative_low
-        {
-            DualAxisData::new(x, y)
-        } else {
-            DualAxisData::new(0.0, 0.0)
+        match dual_axis.deadzone_shape {
+            DeadzoneShape::Cross => {
```
alice-i-cecile, line 390: This logic should all be refactored into a reusable public method or two on `DeadZoneShape` :)

alice-i-cecile:
WRT question 3, I'm increasingly convinced that we probably don't even want to store a `SingleAxis` inside of `DualAxis`: it's overly clever and the abstraction doesn't quite work. @paul-hansen, @Shatur: do either of you have opinions on that (or the rest of this PR)?

Shatur:
Never used this API, so I don't have an opinion on it :(

src/axislike.rs
100-TomatoJuice: I initially did do this like:
```rust
pub enum DeadzoneShape{
    Cross(f32, f32),
}
```
but felt that `SingleAxis` thresholds then served no purpose, so went for using the thresholds. But, I agree with merging in the properties now.

Would the merge look like:
```rust
// 1
pub enum DeadzoneShape{
    Cross(f32, f32),
}

// 2
pub enum DeadzoneShape{
    Cross {x: f32, y: f32},
}

// or 3?
pub enum DeadzoneShape{
    Cross(CrossShapeProperties),
}

pub struct CrossShapeProperties{
    x: f32,
    y: f32
}
```

- 1 is the fine lacks documentation on what the values mean unless I placed the docs at `Cross`
- 2 is more clear than 1 but might be too much
- 3 seems like way too much but is an option and might be the best

src/axislike.rs
100-TomatoJuice: This hard to put into words as it easier to see with an image but, `Cross` is two intersecting rectangles that have can have their widths changed independently. Is that clear, or does it still need some work shopping?

src/axislike.rs
100-TomatoJuice: I agree with removing the circle and adding an ellipse. It provides more control over a circle. It is the same reason I went for using a rectangle over a square.

src/axislike.rs
alice-i-cecile: 2 is my preference: named fields are really nice.

src/axislike.rs
alice-i-cecile: Hmm, still not clear to me. Eventually I'd like to have an interactive example that draws them, but that can wait for a new PR.

src/axislike.rs
```diff
@@ -688,3 +705,105 @@ impl From<DualAxisData> for Vec2 {
...
+
+impl DeadZoneShape {
+    /// 
+    pub fn get_dual_axis_data(
+        &self,
+        dual_axis: &DualAxis,
+        input_stream: &InputStreams,
+    ) -> DualAxisData {
+        match self {
+            DeadZoneShape::Cross {
```
100-TomatoJuice, line 756: Cross is a little odd compared to Rect and Ellipse. Technically the implemention of `Cross` is not really a deadzone, it's more snap zone. Instead of return (0,0) if it's inside the cross, it currently snaps the value to north, south, east, or west depending on where the joystick is pointing. Should I change it to be an actual "deadzone" or keep it as is? I'm in favor of making a deadzone instead, as the current snapping doesn't really turn the joystick into a dpad, only part of it. The corners of the dpad-joystick still have full f32 precision.

src/axislike.rs
```diff
@@ -688,3 +705,105 @@ impl From<DualAxisData> for Vec2 {
...
+        radius_y: f32,
+    },
+}
+
+impl Eq for DeadZoneShape {}
+impl std::hash::Hash for DeadZoneShape {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {}
+}
+
+impl DeadZoneShape {
```
100-TomatoJuice, line 748: @alice-i-cecile 
Is this how you imagined the `impl` to look. Or should `&InputSteams` not be a parameter? Processing the `DualAxis` x and y in `extract_dual_axis_data()` is only really feasible if `Cross` becomes a deadzone instead of a "snapzone", at least from my understanding.

src/axislike.rs
```diff
@@ -688,3 +705,105 @@ impl From<DualAxisData> for Vec2 {
...
+    /// 
+    pub fn get_dual_axis_data(
+        &self,
+        dual_axis: &DualAxis,
+        input_stream: &InputStreams,
+    ) -> DualAxisData {
+        match self {
+            DeadZoneShape::Cross {
+                horizontal_width,
```
100-TomatoJuice, line 757: Also, because `Cross` is more of a "snapzone", it makes `horizontal_width` and `vertical_width` harder to use as it requires setting the `SingleAxis` thresholds.

src/axislike.rs
100-TomatoJuice: Actually, if `Cross` becomes an actual deadzone, then `dual_axis: &DualAxis` and  `input_stream: &InputStreams` both won't be needed to be passed in, and the function can become something like, `point_outside_shape(self, x, y) ` and return a `bool` instead of `DualAxisData`. Then `extract_dual_axis_data()` can get the x, y input values again like:
```rust
fn extract_dual_axis_data(&self, dual_axis: &DualAxis) -> DualAxisData {
    let x = input_stream.input_value(&UserInput::Single(InputKind::SingleAxis(dual_axis.x)), false);
    let y = input_stream.input_value(&UserInput::Single(InputKind::SingleAxis(dual_axis.y)), false);

    if dual_axis.deadzone.point_inside_shape(x, y){
        DualAxisData::new(x, y)
    } else {
        DualAxisData::new(0, 0)
    }
}
```
Which is way cleaner imo. This does have the "loss of functionally" of per-controller snapping, but Bevy kinda does this with livezone in `AxisSettings` in the `GamepadSettings` resource. Plus, if per-controller snapping, instead of global, was needed, it would be quite easy, I think, to add snapping to `DualAxis`.

src/axislike.rs
```diff
@@ -688,3 +703,114 @@ impl From<DualAxisData> for Vec2 {
...
+        /// The horizontal radius of the ellipse.
+        radius_x: f32,
+        /// The vertical radius of the ellipse.
+        radius_y: f32,
+    },
+}
+
+impl Eq for DeadZoneShape {}
+impl std::hash::Hash for DeadZoneShape {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
```
100-TomatoJuice, line 744: I have limited knowledge on how to implement `Hash`, so I just looked at `SingleAxis` and tried to go from there. Is this correct?

src/axislike.rs
```diff
@@ -194,16 +196,27 @@ impl DualAxis {
     /// This cannot be changed, but the struct can be easily manually constructed.
     pub const DEFAULT_DEADZONE: f32 = 0.1;
 
-    /// Creates a [`DualAxis`] with both `positive_low` and `negative_low` in both axes set to `threshold`.
+    /// The default shape of the deadzone used by constructor methods.
+    ///
+    /// This cannot be changed, but the struct can be easily manually constructed.
+    pub const DEFAULT_DEADZONE_SHAPE: DeadZoneShape = DeadZoneShape::Cross {
```
100-TomatoJuice, line 202: Should the default shape be a cross? I think ellipse is a more common, but it's not a big deal either way.

alice-i-cecile: Let's go with a circle.

src/axislike.rs
alice-i-cecile: This seems fine.

No clue about cargo fmt; I'm fine with leaving it a mystery.

alice-i-cecile (approved):

PR #376, opened 2023-08-26
Default controls example

### LINKED ISSUE #248: Demonstrate pattern for exhaustive default controls
Users can conveniently define and then insert default input bindings for their controls by defining methods on their action types which exhaustively match on the action to return a `UserInput`. These can then be used to populate the `InputMap` as follows:

```rust
        let mut input_map = InputMap::default();

        for action in PlayerAction::variants() {
            input_map.insert(PlayerAction::default_keyboard_mouse_input(action), action);
            input_map.insert(PlayerAction::default_gamepad_input(action), action);
        }
```

This should be demonstrated in an example.

100-TomatoJuice:
Fixes #248 

This is my first time creating an example, so I'd be happy to hear any feedback on the comments. Is this what you had in mind for the example?

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -8,6 +8,7 @@
 ### Docs
 
 - Fixed invalid example code in README
+- Added example for setting default controls
 
 ### Enhancements
 - Added `DeadZoneShape` for `DualAxis` which allows for different deadzones shapes: cross, rectangle, and ellipse.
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):
This is perfect: it clearly demonstrates the pattern and the comments are helpful and friendly :)

PR #377, opened 2023-08-29
Add sensitivity

### LINKED ISSUE #322: Allow adding a multiplier for certain inputs' value

When setting an action for both mouse motion and gamepad stick (classic example - camera turning) the gamepad sends values from 0 to 1 while the mouse sends values in pixels (I think?). This means the slightest move of the mouse makes the camera spin orders of magnitude faster than when pushing the thumbstick all the way.

Currently I work around this by using different actions - one for the mouse and one for the gamepad - and combining their result in the input handling system. But it would be nice if I didn't have to split the action based on input source...

Allow setting a factor in the `InputMap` for specific inputs, and have the system that updates the `ActionState` multiply the values by that number.

Since this is mainly a mouse problem, it could also work - and potentially be more efficient - to add a multiplier only for mouse input via a resource.

100-TomatoJuice:
Closes #322 

The design of the how to implement sensitivity was partly discussed in #322. However, I will paste my design comment for easy reading.

## How to add the multiplier to inputs
My design would be something like this:
```rust
input_map: InputMap::default()
    .insert(DualAxis::mouse_motion().with_sensitivity(0.1), Action::Look)
    .insert(DualAxis::left_stick().with_sensitivity(1.5), Action::Look)
```
The value passed in for `with_sensitivity()` would be a multiplier, 1.0 would be the default.

## Where to apply the multiplier
It might be possible to place it in `input_value()` in `input_streams.rs`. It would look something like this:
```rust
let value_in_axis_range = |axis: &SingleAxis, value: f32| -> f32 {
    if value >= axis.negative_low && value <= axis.positive_low && include_deadzone {
        0.0
    } else if axis.inverted {
        -value * axis.sensitivity
    } else {
        value * axis.sensitivity
    }
};
```
Every axis, as far as I know, uses `value_in_axis_range()`, so it might be as easy as just adding `* axis.sensitivity`. Also, from looking around a bit, every input seems to at some point call `input_value()`. So, calculating the sensitivity there will make sure all inputs will be modified if needed.

## Where to store the data
I was thinking we could store the sensitivity value in `SingleAxis` as `DualAxis` is just two `SingleAxis`. Then, we could add constructor methods as shown above for both `SingleAxis` and `DualAxis`.
```rust
// `SingleAxis` verision
#[must_use]
pub fn with_sensitivity(self, sensitivity: f32) -> SingleAxis {
    self.sensitivity = sensitivity;
    self
}

// `DualAxis` verision
#[must_use]
pub fn with_sensitivity(self, y_sensitivity: f32, y_sensitivity: f32) -> SingleAxis {
    self.x.sensitivity = x_sensitivity;
    self.y.sensitivity = y_sensitivity;
    self
}
```
## How to change the data
I think a way to change the sensitivity would be to query for `InputMap` and use `get()`. It does return `&PetitSet<UserInput, 16>` so that may be hard to work with as you have to index into the correct `UserInput`. I don't know if there can even be an easy way to do this. Do users have to look up to order they inserted the `UserInput` into `InputMap` to change the correct one? This is still an open question that I would love to hear about!
```rust
fn set_input_sensitivity(query: Query<&mut InputMap>) {
    let map = query.single();
    // Changing the mouse input sensitivity set in above example
    let input = map.get(Action::Look).get_at_mut(0).unwrap();
    input.sensitivity = 0.4;
}
```

Things left to do:

- [x] Add docs
- [x] Add a doc example
- [x] Add tests
- [x] Update RELEASE.md
- [x] Update `PartialEq` and `Hash` for `SingleAxis`

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -12,6 +12,7 @@
 
 ### Enhancements
 - Added `DeadZoneShape` for `DualAxis` which allows for different deadzones shapes: cross, rectangle, and ellipse.
+- Added sensitivity for `SingleAxis` and `DualAxis`.
 
 ## Version 0.10
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -8,6 +8,7 @@
 ### Docs
 
 - Fixed invalid example code in README
+- Added example for setting default controls
 
 ### Enhancements
 - Added `DeadZoneShape` for `DualAxis` which allows for different deadzones shapes: cross, rectangle, and ellipse.
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -12,7 +12,7 @@
 
 ### Enhancements
 - Added `DeadZoneShape` for `DualAxis` which allows for different deadzones shapes: cross, rectangle, and ellipse.
-- Added sensitivity for `SingleAxis` and `DualAxis`.
+- Added sensitivity for `SingleAxis` and `DualAxis`, allowing you to scale mouse, keypad and gamepad inputs differently for each action.
 
 ## Version 0.10
 
--- a/src/axislike.rs
+++ b/src/axislike.rs
@@ -28,10 +28,11 @@ pub struct SingleAxis {
     pub positive_low: f32,
     /// Any axis value lower than this will trigger the input.
     pub negative_low: f32,
+    /// Whether to invert output values from this axis.
+    pub inverted: bool,
     /// How sensitive the axis is to input values.
     ///
     /// Since sensitivity is a multiplier, any value `>1.0` will increase sensitivity while any value `<1.0` will decrease sensitivity.
-    /// Negative values will invert the axis.
     pub sensitivity: f32,
     /// The target value for this input, used for input mocking.
     ///
@@ -47,6 +48,7 @@ impl SingleAxis {
             axis_type: axis_type.into(),
             positive_low: threshold,
             negative_low: -threshold,
+            inverted: false,
             sensitivity: 1.0,
             value: None,
         }
@@ -62,6 +64,7 @@ impl SingleAxis {
             axis_type: axis_type.into(),
             positive_low: 0.0,
             negative_low: 0.0,
+            inverted: false,
             sensitivity: 1.0,
             value: Some(value),
         }
@@ -74,6 +77,7 @@ impl SingleAxis {
             axis_type: AxisType::MouseWheel(MouseWheelAxisType::X),
             positive_low: 0.,
             negative_low: 0.,
+            inverted: false,
             sensitivity: 1.0,
             value: None,
         }
@@ -86,6 +90,7 @@ impl SingleAxis {
             axis_type: AxisType::MouseWheel(MouseWheelAxisType::Y),
             positive_low: 0.,
             negative_low: 0.,
+            inverted: false,
             sensitivity: 1.0,
             value: None,
         }
@@ -98,6 +103,7 @@ impl SingleAxis {
             axis_type: AxisType::MouseMotion(MouseMotionAxisType::X),
             positive_low: 0.,
             negative_low: 0.,
+            inverted: false,
             sensitivity: 1.0,
             value: None,
         }
@@ -110,6 +116,7 @@ impl SingleAxis {
             axis_type: AxisType::MouseMotion(MouseMotionAxisType::Y),
             positive_low: 0.,
             negative_low: 0.,
+            inverted: false,
             sensitivity: 1.0,
             value: None,
         }
@@ -123,6 +130,7 @@ impl SingleAxis {
             axis_type: axis_type.into(),
             negative_low: threshold,
             positive_low: f32::MAX,
+            inverted: false,
             sensitivity: 1.0,
             value: None,
         }
@@ -136,6 +144,7 @@ impl SingleAxis {
             axis_type: axis_type.into(),
             negative_low: f32::MIN,
             positive_low: threshold,
+            inverted: false,
             sensitivity: 1.0,
             value: None,
         }
@@ -159,7 +168,7 @@ impl SingleAxis {
     /// Returns this [`SingleAxis`] inverted.
     #[must_use]
     pub fn inverted(mut self) -> Self {
-        self.sensitivity = -self.sensitivity;
+        self.inverted = !self.inverted;
         self
     }
 }
--- a/src/axislike.rs
+++ b/src/axislike.rs
@@ -28,8 +28,6 @@ pub struct SingleAxis {
     pub positive_low: f32,
     /// Any axis value lower than this will trigger the input.
     pub negative_low: f32,
-    /// Whether to invert output values from this axis.
-    pub inverted: bool,
     
     pub sensitivity: f32,
     /// The target value for this input, used for input mocking.
@@ -46,7 +44,6 @@ impl SingleAxis {
             axis_type: axis_type.into(),
             positive_low: threshold,
             negative_low: -threshold,
-            inverted: false,
             sensitivity: 1.0,
             value: None,
         }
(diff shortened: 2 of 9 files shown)


100-TomatoJuice:
Continuing from #322 
> A global resource would not allow that.

I don't think it would necessarily disallow a resource, as mouse motion is a `DualAxis` same with a controller stick. But, it would be a lot of work to track all these different inputs. A resource also creates the question of when to add to this resource? Making the users add to that even if that don't care about sensitivity seems a bit verbose. Maybe using a `Query` would be the best option? As, it allows for filtering to get the correct player. Whatever the end result for changing sensitivity, I would like to create an example to showcase how to do it.

100-TomatoJuice:
Is there any way to tell the difference from, say, a controller stick and mouse motion `DualAxis`? I was thinking of using a `HashMap` with a `InputKind` as a key and a `UserInput` as a value, but I don't know if that would work, or even be any better than the current running implementation. The solution will need access to `InputMap` as that is what is needed. The question is how to cleanly extract the needed `UserInput` from `map`?

100-TomatoJuice:
Changing the code 
```rust 
// This actually only gets the `UserInput` so there is quite a bit of matching needed to get to a Dual/Single Axis
let input = map.get(Action::Look).get_at_mut(0).unwrap();
```
with 
```rust
let input = map.get_with_type(Action::Look, LeftStick)
``` 
seems really nice, but I can't think of a way to say "I want the left stick for this action" or "I want the mouse motion for this action". The closest to this is `InputKind`, but I don't know if it is enough. The line of thought here is a user would choose what action and what input to get the `DualAxis`, `SingleAxis`, or whatever input they want. This could also be used to change other things like the deadzone as well.

100-TomatoJuice:
After a bit of testing, this is the current way to change input from an `InputMap`
```rust
fn change_input_values(mut query: Query<&mut InputMap<Action>, With<Player>>){
    let mut input_map = query.single_mut();

    let input = input_map.get_mut(Action::Look).get_at_mut(0).unwrap();
    if let UserInput::Single(kind) = input{
        if let InputKind::DualAxis(dual_axis) = kind{
            println!("X: {}, Y: {}, Shape: {:?}", dual_axis.x.sensitivity, dual_axis.y.sensitivity, dual_axis.deadzone);
            dual_axis.x.sensitivity = 0.8;
            dual_axis.y.sensitivity = 0.8;
            dual_axis.deadzone = DeadZoneShape::Rect { width: 1.0, height: 1.0 }
        }
    }
}
```
It is a little on the gross side, but I can't really think of a good way to select with input you want to modify. The first parameter is `Actionlike`, but the how to remove the `get_at_mut(0)` and the pattern matching is the next question. Passing in a `UserInput` like `UserInput::Single(InputKind::DualAxis( **There is nothing to pass in here**  ))` won't work.

Is doing this pattern matching the only way? If it is, then an example, which I could make for this pr, is definitely needed as changing values is verbose.

src/axislike.rs
```diff
@@ -30,6 +30,8 @@ pub struct SingleAxis {
     pub negative_low: f32,
     /// Whether to invert output values from this axis.
     pub inverted: bool,
+    
+    pub sensitivity: f32,
```
alice-i-cecile, line 34: Precise doc strings will be important here.

src/axislike.rs
```diff
@@ -30,6 +30,8 @@ pub struct SingleAxis {
     pub negative_low: f32,
     /// Whether to invert output values from this axis.
     pub inverted: bool,
```
alice-i-cecile, line 32: We can replace `inverted` with a negative sensitivity value.

src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![forbid(missing_docs)]
+//#![forbid(missing_docs)]
```
alice-i-cecile, line 1: Remember to revert this :)

alice-i-cecile (changes requested):
Alright, that's more straightforward than I expected. I'm not *super* concerned about being able to tweak these values later, the fact that it's possible to at all is about good enough for me.

We should collapse the `inverted` field into `sensitivity`, write up a change log and add some tests before merging and shipping this though.

src/axislike.rs
plof27: We could do this mathematically, but I think it makes the API a lot less clear. I think we should keep them separate, and add some kind of non-negative constraint on sensitivity (or at least a warning).

alice-i-cecile: Okay, fair. Let's do that then.

100-TomatoJuice:
Ok, I will get on that and all the changes presented. Do you want a new example for how to change values at runtime, punt it to a new pr, or just not have an example?

src/axislike.rs
100-TomatoJuice: Sorry, but what does "that" refer to? I'm guessing merging sensitivity and inverted as you said below, but what do about making that making the API less clear as plof27 said. I'm guessing just write useful docs on sensitivity to explain how inverting works now?

100-TomatoJuice: As a side note, instead of removing the invert functions, I can just invert the sensitivity to do the same thing.
```rust
/// Returns this [`SingleAxis`] inverted.
#[must_use]
pub fn inverted(mut self) -> Self {
    self.sensitivity = -self.sensitivity;
    self
}
```

alice-i-cecile:
@plof27 convinced me: let's revert the change to merge inverted + sensitivity. It's technically correct, but a less clear API.

Add `get_mut` in this PR, and make a doc test demonstrating how to change the values :)

100-TomatoJuice:
Should I add  "some kind of non-negative constraint on sensitivity (or at least a warning)" as plof27 said? If so, where would that go, as there a many ways of creating an axis. Should I add a doc on sensitivity that negative values will invert it, so keep the values `>=0.0`?

RELEASES.md
```diff
@@ -12,6 +12,7 @@
 
 ### Enhancements
 - Added `DeadZoneShape` for `DualAxis` which allows for different deadzones shapes: cross, rectangle, and ellipse.
+- Added sensitivity for `SingleAxis` and `DualAxis`.
```
alice-i-cecile, line 15:
```suggestion
- Added sensitivity for `SingleAxis` and `DualAxis`, allowing you to scale mouse, keypad and gamepad inputs differently for each action.
```

alice-i-cecile:
> Should I add "some kind of non-negative constraint on sensitivity (or at least a warning)" as plof27 said

I'm not fully sure where this should go. We *could* put it on every constructor / setter? But that's pretty duplicative, and would force us to fully lock down the construction.

We could also use a custom `StrictlyPositivef32` type or something, but the ergonomics of working with those is very frustrating IME.

I'm tempted to just add it to the docs for now and get this merged and shipped.

plof27:
Agreed

src/axislike.rs
100-TomatoJuice: > Ah, I like that :)
So, would you like that re-merged into one value? Or, close this as resolved and keep them separate. I am good with either solution.

src/input_map.rs
```diff
@@ -19,60 +19,101 @@ use std::collections::HashMap;
...
+        if let InputKind::DualAxis(dual_axis) = kind{
+            // Here any value of the left stick `DualAxis` can be changed
+            dual_axis.x.sensitivity = 0.8;
+            dual_axis.y.sensitivity = 0.8;
+            dual_axis.deadzone = DeadZoneShape::Rect { width: 1.0, height: 1.0 }
+        }
+    }
+}
+```
+**/
```
100-TomatoJuice, line 116: Should these `**/` lines be removed? I think vscode added them for some reason.

alice-i-cecile: This is correct: it's to replace a `///` doc comment indicator on each line :)

alice-i-cecile (approved):

PR #385, opened 2023-09-12
Fix compile error when default features are off

### LINKED ISSUE #383: Compile error with default features turned off
## Version
Main

## What you did

Turned off default features for LWIM
```
[dependencies]
bevy = { version = "0.11.2", default-features = false}
leafwing-input-manager = { git = "https://github.com/Leafwing-Studios/leafwing-input-manager.git", default-features = false}
```

The window to be created and the app run normally

## What actually happened

A compile error is printed in the terminal

## Additional information
Log:
```
failed to resolve: use of undeclared type `UiSystem`
   --> leafwing-input-manager/src/plugin.rs:118:32
    |
118 |                         .after(UiSystem::Focus)
```

This error is introduced in #375.

100-TomatoJuice:
Closes #383 
I tested minimal, ui_driven_actions, and bindings_menu to test different features. They all worked as expected.

### DIFF
--- a/src/plugin.rs
+++ b/src/plugin.rs
@@ -100,31 +100,27 @@ impl<A: Actionlike> Plugin for InputManagerPlugin<A> {
                 )
                 .add_systems(PostUpdate, release_on_input_map_removed::<A>);
 
-                #[cfg(feature = "egui")]
-                app.add_systems(
+                app.configure_set(
                     PreUpdate,
-                    update_action_state::<A>
+                    InputManagerSystem::Update
                         .run_if(run_if_enabled::<A>)
-                        .in_set(InputManagerSystem::Update)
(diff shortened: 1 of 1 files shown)


src/plugin.rs
```diff
@@ -133,6 +129,18 @@ impl<A: Actionlike> Plugin for InputManagerPlugin<A> {
                         .after(UiSystem::Focus)
                         .after(InputSystem),
                 );
+
+                app.add_systems(
```
alice-i-cecile, line 133: I think this will be cleaner if we move it above the feature-gated code.

100-TomatoJuice: I moved it above the `configue_sets()` to pair it with the other `add_systems()`

PR #386, opened 2023-09-15
Implement Reflect for InputMap

barsoosayque:
Currently it's not possible to directly use an `InputMap<A>` as an asset (e.g. `Handle<InputMap<A>>`), as `trait Asset` now requires `TypePath`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,9 @@ bevy_egui = { version = "0.22", optional = true }
 
 petitset = { version = "0.2.1", features = ["serde_compat"] }
 derive_more = { version = "0.99", default-features = false, features = [
+  "as_ref",
+  "as_mut",
+  "from",
   "display",
   "error",
 ] }
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -16,6 +16,7 @@
 
 ### Enhancements
 
+- Implement `TypePath` for `InputMap`.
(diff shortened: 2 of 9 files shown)


alice-i-cecile:
Can you swap this to a `Reflect` impl instead? It will have the same effect, but is much more powerful.

barsoosayque:
Sure thing. Didn't know `Reflect` also implements `TypePath`, but I'll try to derive it for other types too

barsoosayque:
So since `PetitSet` is a foreign type, I had to create a newtype for it just to have a reflectable `PetitSet`, and then manually implement `Reflect` (and other reflect related traits). It doesn't seem like a bad solution, but one major issue is that `UserInput::Chord` is not a `Box` anymore (since `Box<dyn Reflect>` is not reflectable, [yet](https://github.com/bevyengine/bevy/issues/3392)), therefore it drastically increased in size and I'm not really sure what to do about it.

alice-i-cecile:
Right, okay :) Let's merge this as is then.

barsoosayque:
Um. so this PR is pretty much the same as #331 (although it's slightly outdated). And I've seen that there is an ongoing work to remove a `PetitSet` as a dependency, so I'm not sure what's the best course of action here..

alice-i-cecile:
We should merge this and unblock this feature request, then remove PetitSet in another PR.

alice-i-cecile:
I'm going to tackle this in #400, which should be done shortly :) This was super helpful to push me towards "remove petitset" as the right architectural decision, thank you.

PR #387, opened 2023-09-16
Add gamepad registration example

### LINKED ISSUE #18: Create a gamepad registration example
Use a Nintendo Switch style "press L and R" to register prompt.

100-TomatoJuice:
Fixes #18

Teaches how to use LWIM for joining with gamepads in a local multiplayer fashion.

### DIFF
--- a/examples/register_gamepads.rs
+++ b/examples/register_gamepads.rs
@@ -1,4 +1,5 @@
 //! Demonstrates how to register gamepads in local multiplayer fashion
+
 use bevy::{prelude::*, utils::HashMap};
 use leafwing_input_manager::prelude::*;
 
@@ -55,6 +56,8 @@ fn join(
                     input_map: InputMap::default()
                         .insert(GamepadButtonType::South, Action::Jump)
                         .insert(GamepadButtonType::Select, Action::Disconnect)
+                        // Make sure to set the gamepad or all gamepads will be used!
+                        .set_gamepad(gamepad)
                         .build(),
                 })
                 .insert(Player { gamepad })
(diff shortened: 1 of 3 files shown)


examples/register_gamepads.rs
```diff
@@ -0,0 +1,98 @@
...
+fn join(
+    mut commands: Commands,
+    mut joined_players: ResMut<JoinedPlayers>,
+    gamepads: Res<Gamepads>,
+    button_inputs: Res<Input<GamepadButton>>,
+) {
+    for gamepad in gamepads.iter() {
+        // Join the game when both bumpers (L+R) on the controller are pressed
+        // We drop down the Bevy's input to get the input from each gamepad
+        if button_inputs.just_pressed(GamepadButton::new(gamepad, GamepadButtonType::LeftTrigger))
```
100-TomatoJuice, line 40: This is kinda unreliable in terms of if is works or not. Sometimes I had to press (L+R) several times before it joined. Using `pressed()` works, but it fills the terminal with "Player 0 already joined!" for a second, and it turns the behavior of "join when L and R are pressed together" to "join when L and R are both pressed, regardless of if it was done together". If this behavior is acceptable, I can make the change and could remove the print for a player joining when they are already joined.

alice-i-cecile: We should swap these both to `pressed`, not `just_pressed`. That's both more reasonable (why do we need a frame-perfect input to join the game?) and fix the reliability.

examples/register_gamepads.rs
```diff
@@ -0,0 +1,98 @@
...
+            && button_inputs
+                .just_pressed(GamepadButton::new(gamepad, GamepadButtonType::RightTrigger))
+        {
+            // Make sure a player can not join twice
+            if joined_players.0.contains_key(&gamepad) {
```
alice-i-cecile, line 45: I think we should invert this logic and then skip the warning message.

alice-i-cecile (approved):

PR #355, opened 2023-06-24
Do not trigger mouse actions on button click

Shatur:

PR #391, opened 2023-10-07
Add `block_ui_interactions` feature

100-TomatoJuice:
Adds the `block_ui_interactions` feature, which allows users to optionally disable #355. 

I chose to place the feature in the default feature to keep the current behavior. As for the feature itself, I chose not to turn on `ui` or `bevy_ui`. This makes users explicitly state that `ui` is on, and they want to block ui interaction as well. That does mean that turning on `block_ui_interactions` will do nothing, as it relies on `ui`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,9 @@ opt-level = 3
 members = ["./", "tools/ci", "macros"]
 
 [features]
-default = ['ui']
+default = ['ui', 'block_ui_interactions']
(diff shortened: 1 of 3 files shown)


100-TomatoJuice:
Is that error because of Rust 1.73?

100-TomatoJuice:
Error looks to be due to Rust 1.73

alice-i-cecile (approved):

PR #392, opened 2023-10-07
Fix `redundant explicit link target` docs error

100-TomatoJuice:
Fix doc error due to `redundant explicit link target` that was introduced in Rust 1.73


alice-i-cecile:
Thanks!

PR #395, opened 2023-10-11
Fix divide by zero in ellipse deadzone

100-TomatoJuice:
As vero brought up in Discord, when either ellipse radius is 0, `NaN` is returned. So to fit with the other shapes when they are 0, if either radius is 0 then return true to allow any input value to be read.

### DIFF
--- a/src/axislike.rs
+++ b/src/axislike.rs
@@ -842,6 +842,10 @@ impl DeadZoneShape {
 
     /// Returns whether the (x, y) input is outside an ellipse.
     fn outside_ellipse(&self, x: f32, y: f32, radius_x: f32, radius_y: f32) -> bool {
+        if radius_x == 0.0 || radius_y == 0.0 {
+            return true;
+        }
+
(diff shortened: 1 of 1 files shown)


alice-i-cecile (changes requested):
Changelog entry please :) Then this LGTM and I'll merge.

100-TomatoJuice:
Done :). But, does it make sense to add a bug fix to a changelog to an unreleased feature?

alice-i-cecile:
Oh right lol. Yeah, we should revert that.

RELEASES.md
```diff
@@ -7,6 +7,7 @@
 - Fixed system order ambiguity between bevy_ui and update_action_state systems
 - The input values of axis inputs in a `Chord` are now prioritized over buttons
 - Fixed unassigned `InputMaps`s not receiving input from all connected gamepads
+- Fixed ellipse deadzone shape dividing by zero when radius is zero
```
alice-i-cecile, line 10:
```suggestion
```

alice-i-cecile (approved):

PR #396, opened 2023-10-11
Change deadzone to have boundary mean outside shape

100-TomatoJuice:
Having the boundary of the `DeadZoneShape` mean that it is outside is in line with #395. This also means that any `DeadZoneShape` that has a volume of 0 will let all input values be read.


alice-i-cecile:
Can you add a quick set of unit tests for this please?

100-TomatoJuice:
Ya, I add a test to make sure all (0, 0) with a 0 volume get the input

PR #409, opened 2023-11-05
Add back `bevy_egui` to LWIM 0.11

100-TomatoJuice:
I tested `minimal.rs` using the egui feature just to make sure nothing broke, and it works.


alice-i-cecile (changes requested):
Can you update the release notes and the main Cargo.toml to a point release here too?

100-TomatoJuice:
Like changing the LWIM version to 0.11.1 and adding to the release notes that egui was added back?

alice-i-cecile:
Correct. I don't bother with the minor versions for the macro crate.

alice-i-cecile (approved):

PR #415, opened 2023-11-13
Add support for more complex Action types

### LINKED ISSUE #68: Allow users to store ActionStates for action types that cannot be compile-time enumerated
Ideally, we can store a trait object to allow for easy and powerful extension.

This is most important for RPGs, who have a very, very large list of actions, or for modding support.

Typically, this would be combined using the two layers of actions + indirection pattern here, or via the UI: https://github.com/Leafwing-Studios/leafwing-input-manager/blob/main/examples/arpg_indirection.rs

### LINKED ISSUE #278: Handling of enum with sub-variants
Hi,

I am not sure if this is a bug/limitation of the current implementation but I encounter some issues with the Actionlike macro being used with "complex" enums: it seems to construct them by default, thus loosing the information.

```rust
#[derive(PartialEq, Default, Eq, Clone, Debug, Hash, Copy)]
enum ActionBars { #[default] AB0, AB1 }

#[derive(Actionlike, PartialEq, Eq, Clone, Debug, Copy)]
enum GameAction { Jump, ActionBar(ActionBars), Test(i32) }
```

First I was "forced" to add the Default trait to my `ActionBars` enum, but when running, it seems that everytime an `ActionBar` or the `Test` variant of the enum was triggered, it was reset to default when passed to the `ActionState`.

Code that spawns the bundle
```rust
commands.spawn_bundle(PlayerInputBundle{
        player: Player,
        input_map: InputMap::new([
            (KeyCode::Key1, GameAction::ActionBar(ActionBars::AB0)),
            (KeyCode::Key2, GameAction::ActionBar(ActionBars::AB1)),
            (KeyCode::E, GameAction::Test(42)),
            (KeyCode::Space, GameAction::Jump)
        ]).build(),
        ability_action_state: ActionState::default(),
        actionbar_map: actionbars_mapping
    });
```

Actually reading the action:
```rust
fn print_abilities(query: Query<(&ActionState<GameAction>, &ActionBarsMap)>) {
    for (ability_state, mapping) in query.iter() {
        for ability in ability_state.get_just_pressed() {
            if let GameAction::ActionBar(ab) = ability {
                dbg!(ab);
                let spell_id = mapping.get(&ab);
                dbg!(spell_id);
            } else {
                dbg!(ability);
            }
        }
    }
}
```
And in the case of subtypes, it always prints either `ActionBars(AB0)` or `Test(0)`, even when it should have printed `ActionBars(AB1)` or `Test(42)`.

Is this a limitation or did I miss something to make this work?

If it is a limitation, I would like to have this as an enhancement if possible.

Thank you

### LINKED ISSUE #412: Remove `variants` and related methods from `Actionlike`
There are still quite a bit of `A::variants()` usage left. I'm kinda surprised that it has not broken anything else like it did. Is it within the scope of this pr to remove usage of `A::variants()` as you mentioned on Discord? The removal of `A::variants()` will not be trivial, I think. For `InputMap` it can use `iter()`, but `ActionState` doesn't have access to the map so it will require some thought.

_Originally posted by @100-TomatoJuice in https://github.com/Leafwing-Studios/leafwing-input-manager/issues/400#issuecomment-1806846091_

alice-i-cecile:
This is a complex PR: it completely refactors our internals in order to allow support for more complex action types and generally reduce internal jank.

Namely, it:

1. Removes `Actionlike::variants` and other methods that assume a finite ordered list of discrete actions.
2. Passes action types by reference rather than value.
3. Removes the existing `DynamicAction` types.

Fixes #412. Fixes #278. Fixes #68.

# Remaining items

- [ ] fix tests
- [ ] fix doc tests
- [ ] check examples


alice-i-cecile:
Closing this and doing this in a more incremental way.

PR #421, opened 2023-12-02
Fix `egui` feature compilation error 

devil-ira:
Fixes a compile error with the `egui` feature enabled when the `ui` or `block_ui_interactions` features are disabled.


WesleyClements:
I've looked into the tests failing a bit and it seems that changing the mouse_motion to a Vec is preventing app.send_input from doing anything. Specifically, the MouseMotion is being created but not propagated back into update_action_state because self.mouse_motion.push is a mutation on a copy of the data.

WesleyClements:
I addressed the failing tests in #424 but do make sure that the changes I made make sense

alice-i-cecile:
You're both great: thank you. I'm currently sick, but I'm looking to sit down with LWIM later this week and fix stuff up.

alice-i-cecile (approved):

PR #422, opened 2023-12-06
Add ValueChange and AxisPairChange variants to ActionDiff

### LINKED ISSUE #342: Axis values support for network diffs
We used the network diffing functionality of this library for our game jam game: https://prawl.itch.io/stellar-squeezebox which overall turned out great and it was really helpful for getting something off the ground quick!

Unfortunately, gamepad support was really lacking when diffing actions with a couple major issues:

A. There is no support for axis data for ActionDiff
B. ~~If you use an axis as a button (e.g. turn left) you can't set a deadzone or it stops being recognized on the server. Example:
  `SingleAxis::negative_only(GamepadAxisType::LeftStickX, 0.1).with_deadzone(deadzone)`~~ Was user error. See [comment](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/342#issuecomment-1510452715)
  
I'd like to contribute towards this, so if you have any thoughts about how you want this to work let me know. (I'll probably taking a few days to relax now that the jam is over, so might be a bit)

WesleyClements:
In this PR I added an `ActionDiff::AxisChanged` variant. This is meant to allow for transmitting changes to axis input types across the network. I've updated the relevant systems to handle creating events for this variant on the client and consuming them on the server. I've been using this solution in my own projects for a little bit and it seems to be working but I'm new to rust in addition to the bevy ecosystem so would really love some feedback as to if this seems like a reasonable solution. There are three potentially breaking changes. First I had to remove `Eq` and `Hash` from ActionDiff as f32 doesn't implement them.  Second I had to add a local resource to the `generate_action_diffs` system to keep track of if the axis actually changed. Lastly I had to add a requirement that the `ID` type implement the `Hash` trait as it is being used a key for a hashmap.

Fixes #342

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,9 @@
 # Release Notes
 
+## Version 0.11.3
+
+- added support in `ActionDiff` for value and axis_pair changes
+
 ## Version 0.11.2
 
 - fixed [a bug](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/285) with mouse motion and mouse wheel events being improperly counted
--- a/examples/send_actions_over_network.rs
+++ b/examples/send_actions_over_network.rs
@@ -24,7 +24,7 @@ enum FpsAction {
 }
 
 /// This identifier uniquely identifies entities across the network
-#[derive(Component, Clone, PartialEq, Eq, Debug)]
+#[derive(Component, Clone, PartialEq, Eq, Hash, Debug)]
 struct StableId(u64);
 
 fn main() {
--- a/src/action_state.rs
+++ b/src/action_state.rs
@@ -4,6 +4,7 @@ use crate::Actionlike;
 use crate::{axislike::DualAxisData, buttonlike::ButtonState};
 
 use bevy::ecs::{component::Component, entity::Entity};
+use bevy::math::Vec2;
 use bevy::prelude::{Event, Resource};
 use bevy::reflect::Reflect;
 use bevy::utils::hashbrown::hash_set::Iter;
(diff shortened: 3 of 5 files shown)


WesleyClements:
I know this will add an extra minimum necessary bit to distinguish between the variants which can potentially add up especially if sending in bulk/with a lot of players. I've been using bincode for serialization which (for the way I have it configured at least 🤷) seems to use at least a byte for the variant signifier. This seems to be the most ergonomic approach as it makes it simpler for the user to only have to handle passing/serializing/deserializing one event type between client and server. That being said, an alternative to this might be to create a separate struct for the AxisDiff and send a different event for those changes. That would force the user to handle both events and figure out how to differentiate between them across the network. I'm using bevy_renet and it wouldn't be that difficult to create a second channel to send the AxisDiffs on. That being said, I'm not sure I would actually reduce the data being sent over the network. I could imagine that the client is typically only changing 8 inputs or so each update which would mean that the extra variant would tacking on an extra byte per request. I'm making another assumption that the channel overhead is comparable to a byte per request so would about the same. I could be very off here and definitely would like feedback but those are my follow up thought after sleeping on this PR and panicking that I hadn't thought through it thoroughly enough 😁.

WesleyClements:
I didn't think it through enough. I realize this doesn't handle single axis input types. I will think of a solution to this problem and implement it real quickly.

WesleyClements:
I added support for syncing `ActionData.value` by including a fourth variant. I also had to refactor the `generate_action_diffs`. It would have been easier if if I knew how to know the input type of an action but I just assumed that if the value of an action is 1.0 that it is a binary input.

RELEASES.md
```diff
@@ -1,5 +1,9 @@
 # Release Notes
 
+## Version 0.11.3
```
WesleyClements, line 3: I didn't know what to do with the versions exactly but incremented the minor at least.

examples/send_actions_over_network.rs
```diff
@@ -24,7 +24,7 @@ enum FpsAction {
 }
 
 /// This identifier uniquely identifies entities across the network
-#[derive(Component, Clone, PartialEq, Eq, Debug)]
+#[derive(Component, Clone, PartialEq, Eq, Hash, Debug)]
```
WesleyClements, line 27: Needed by `generate_action_diffs`

alice-i-cecile:
Good, this seems like a sensible approach! Could you add some more tests to help us spot regressions?

WesleyClements:
This is now dependent on #424

PR #426, opened 2023-12-11
Derive Reflect and Asset for InputMap<A>

### LINKED ISSUE #423: Unable to use InputMap as an asset
## Version

[0.11.2](https://github.com/Leafwing-Studios/leafwing-input-manager/commit/1617f40782323838e7e951a5e2234e6285ed8132)

## Operating system & version

NixOS (master)

## What you did

Updating to bevy 0.12 / leafwing-input-manager 0.11.2

`Handle<InputMap<A>>` generates no errors, as it was valid before to use `InputMap` as an asset.

## What actually happened

My code that used `InputMap<A>` as an asset (e.g. `Handle<InputMap<A>>`) is not working anymore.

## Additional information

Apparently assets now should derive `Asset` instead of `TypeUuid`, and although it is a simple change, it unravels an issue I tried to fix in https://github.com/Leafwing-Studios/leafwing-input-manager/pull/386, because `Asset` requires `TypePath`, which can be provided by `Reflect`. And even though version 0.11 mentions that `InputMap` should be `Reflect` now, as far as I can tell it's not ?

barsoosayque:
This is a continuation from #386. The code is significantly simpler due to how easy it is to omit the multimap dependency so it is possible to derive `Reflect` instead of making a new-type for multimap.

Closes #423.


alice-i-cecile (approved):

alice-i-cecile:
Great work: good call to just swap over to a `HashMap`.

barsoosayque:
CI fails but as far as I can tell it's not because of my changes ?

alice-i-cecile:
Yeah, I'll take a look.

PR #427, opened 2023-12-12
Add `InputMap::clear`

Shatur:
Sometimes it's useful to reset all mappings and re-populate it from scratch. Instead of creating a new `InputMap`, it's better to  reuse the allocated memory.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,9 @@
 # Release Notes
 
+## Unreleased
+
+- Added `InputMap::Clear`.
+
 ## Version 0.11.2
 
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):
Perfectly sensible, thanks!

RELEASES.md
```diff
@@ -1,5 +1,9 @@
 # Release Notes
 
+## Unreleased
+
+- Added `InputMap::Clear`.
```
alice-i-cecile, line 5:
```suggestion
- Added `InputMap::clear`, which works like the `HashMap` equivalent.

```

PR #432, opened 2023-12-28
Fix incorrect axis data in `Chord` when not all buttons are pressed

### LINKED ISSUE #430: Two inputs (WASD, Mouse button + Mouse Movement) that trigger an action, mouse movement will ignore the chord
## Version

0.11.2

## Operating system & version

macOS 14.1.2

## What you did

I have an InputMap that goes into a resource:

```rust
 InputMap::new([
            (UserInput::VirtualDPad(VirtualDPad::wasd()), Self::Move),
            (
                UserInput::Chord(vec![
                    InputKind::Mouse(MouseButton::Middle),
                    InputKind::DualAxis(mouse_motion()),
                ]),
                Self::Move,
            ),
            // other inputs...
        ])
```

Usage in the system:

```rust
fn camera_movement(
    time: Res<Time<Real>>,
    mut action_state: ResMut<ActionState<PlayerAction>>,
    mut camera: Query<(&mut Transform, &OrthographicProjection), With<Camera>>,
) {
    if !action_state.pressed(PlayerAction::Move) {
        return;
    }

    let Some(axis) = action_state.axis_pair(PlayerAction::Move) else {
        return;
    };

    let (mut camera_transform, projection) = camera.single_mut();

    let velocity = axis.xy() * projection.scale;
    camera_transform.translation += velocity.extend(0.);
}

```

These two work as expected separately:
1) WASD will move the camera.
2) Holding down middle mouse button while moving the mouse will move the camera.

But when I hold down one of the WASD keys and move the mouse...

Moving the mouse while pressing any WASD keys should ignore the mouse.

## What actually happened

When I:
1) Hold down one of the WASD keys and
2) Move the mouse and
3) Don't hold down the middle mouse button

... it activates the mouse movement.

Dzjudieng:
Fixes #430

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -6,6 +6,7 @@
 - Register types in the reflection system.
 - added support in `ActionDiff` for value and axis_pair changes
 - Added `InputMap::Clear`.
+- Fixed [a bug](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/407) related to incorrect axis data in `Chord` when not all buttons are pressed.
 
 ## Version 0.11.2
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -6,7 +6,7 @@
 - Register types in the reflection system.
 - added support in `ActionDiff` for value and axis_pair changes
 - Added `InputMap::Clear`.
-- Fixed [a bug](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/407) related to incorrect axis data in `Chord` when not all buttons are pressed.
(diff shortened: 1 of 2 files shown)


alice-i-cecile (changes requested):
Looks great! Can you add an entry to RELEASES.md?

Dzjudieng:
Is there anything else that needs to be changed here?

alice-i-cecile:
Nope, looks great! Thanks a ton.

PR #434, opened 2024-01-02
Update action-diff-system docstring

cBournhonesque:
Tiny docstring change


alice-i-cecile (approved):
Thank you, that's clearly correct <3

PR #435, opened 2024-01-04
Update action-diff to not include the stable component ID

cBournhonesque:
This PR reworks how networking via `ActionDiff` works:
- instead of using a "Stable `ID` Component" that uniquely identifies entities across client and server, we introduce the `ActionDiffEvent` type, which has:
  - an owner: the Entity that generates the action, or None if we are using a global `ActionState`
  - the action_diff itself
- we remove the `process_action_diffs` system in favor of a method `ActionState.apply_diff(action_diff)`, to leave more flexibility to users over how to process the action-diffs (perform entity-mapping, etc.)

I'm slightly confused about this: https://github.com/Leafwing-Studios/leafwing-input-manager/blob/main/src/systems.rs#L216: we detect if the button is Pressed/ValueChanged just based on the value?
What if we have a 'Value' type button that has a value of 1.0? We would then get a `Pressed` `ActionDiff`, which seems incompatible?

NOTE: it might actually be more efficient to have the ActionDiffEvent contains a `Vec<ActionDiff>` since we can compute all of them in one go for a given entity

### DIFF
--- a/src/action_state.rs
+++ b/src/action_state.rs
@@ -501,6 +501,44 @@ impl<A: Actionlike> ActionState<A> {
     pub fn previous_duration(&self, action: A) -> Duration {
         self.action_data[action.index()].timing.previous_duration
     }
+
+    /// Applies an [`ActionDiff`] (usually received over the network) to the [`ActionState`].
+    ///
+    /// This lets you reconstruct an [`ActionState`] from a stream of [`ActionDiff`]s
+    pub fn apply_diff(&mut self, action_diff: &ActionDiff<A>) {
+        match action_diff {
+            ActionDiff::Pressed {
+                action,
+            } => {
+                self.press(action.clone());
+                self.action_data_mut(action.clone()).value = 1.;
+            }
+            ActionDiff::Released {
+                action,
+            } => {
+                self.release(action.clone());
+                let action_data = self.action_data_mut(action.clone());
+                action_data.value = 0.;
+                action_data.axis_pair = None;
+            }
+            ActionDiff::ValueChanged {
+                action,
+                value,
+            } => {
+                self.press(action.clone());
+                self.action_data_mut(action.clone()).value = *value;
+            }
+            ActionDiff::AxisPairChanged {
+                action,
+                axis_pair,
+            } => {
+                self.press(action.clone());
+                let action_data = self.action_data_mut(action.clone());
+                action_data.axis_pair = Some(DualAxisData::from_xy(*axis_pair));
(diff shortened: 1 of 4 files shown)


src/action_state.rs
```diff
@@ -752,44 +780,48 @@ impl Timing {
...
 ///
-/// These are typically accessed using the `Events<ActionDiff>` resource.
+/// These are typically accessed using the `Events<ActionDiffEvent>` resource.
 /// Uses a minimal storage format, in order to facilitate transport over the network.
 ///
-/// `ID` should be a component type that stores a unique stable identifier for the entity
-/// that stores the corresponding [`ActionState`].
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, Event)]
-pub enum ActionDiff<A: Actionlike, ID: Eq + Clone + Component> {
+/// An `ActionState` can be fully reconstructed from a stream of `ActionDiff`
```
alice-i-cecile, line 801:
```suggestion
/// An `ActionState` can be fully reconstructed from a stream of `ActionDiff`.
```

alice-i-cecile:
> I'm slightly confused about this: https://github.com/Leafwing-Studios/leafwing-input-manager/blob/main/src/systems.rs#L216: we detect if the button is Pressed/ValueChanged just based on the value?
What if we have a 'Value' type button that has a value of 1.0? We would then get a Pressed ActionDiff, which seems incompatible?

Yeah, this seems a bit sloppy to me too. Can you make an issue about this?

> NOTE: it might actually be more efficient to have the ActionDiffEvent contains a Vec<ActionDiff> since we can compute all of them in one go for a given entity

Agreed, I think this is important given the networking-centric nature of this code. Feel free to make that change in this PR or a follow-up.

cBournhonesque:
> > I'm slightly confused about this: https://github.com/Leafwing-Studios/leafwing-input-manager/blob/main/src/systems.rs#L216: we detect if the button is Pressed/ValueChanged just based on the value?
> > What if we have a 'Value' type button that has a value of 1.0? We would then get a Pressed ActionDiff, which seems incompatible?
> 
> Yeah, this seems a bit sloppy to me too. Can you make an issue about this?
> 
> > NOTE: it might actually be more efficient to have the ActionDiffEvent contains a Vec since we can compute all of them in one go for a given entity
> 
> Agreed, I think this is important given the networking-centric nature of this code. Feel free to make that change in this PR or a follow-up.

I made the change for the Event to contain a `Vec<ActionDiff>`. 
And sure i'll create an issue!

PR #438, opened 2024-01-05
Accessible Deadzones

### LINKED ISSUE #398: Accessible DeadZone input handling

DeadZoneShape is not accessibility friendly

This discussion is a good read: https://github.com/godotengine/godot-docs/issues/5378#issuecomment-965734470

my conclusions:
there should be two primary modes: circle, and cross. Circle has no amount of axis snapping, because sometimes its not desirable, and cross has some amount of axis snapping. Both circle and cross do range shifting such that all possible inputs are reachable and continuous, (meaning, a small change in input vector causes at most a proportionally small change in output vector. all current deadzones in LWIM do not do this, because crossing the deadzone threshold snaps from 0 to whatever the value is outside the deadzone).

Single axis: `if input.abs() < deadzone { None } else { Some(input.sign() * (input.abs() - deadzone)) }`
Circle: `if input.magnitude() < deadzone { None } else { Some(input.normalized() * (input.magnitude() - deadzone)) }`
Cross: handled like single-axis for each axis individually

## [Optional] How could this be implemented?

Allow deadzone shapes to mutate the passed-in input by returning an `Option<Vec2>` instead of a `bool`. This allows input to be continuous and have unreachable input range, by rescaling/shifting live areas to be zero input on deadzone edges as follows:
141191045-5b6520c5-980e-4df7-917c-d834a6c64a72

This should also be implemented for single axis deadzones, otherwise they will have small unreachable input ranges near zero

- Make users do deadzone handling
this is not ideal because games should be accessible by default. an engine should strive to make it easy for users to get things right, even if they don't know what they are doing.

- Axis snapping decoupled from `DeadZoneShape`
different shapes have different sizes, making snap be a distinct functionality would make it hard to maintain input continuity.

- forget about `DeadZoneShape` and just do single axis deadzones with range shifting on both axes
this works, but we lose the ability to choose different deadzone behaviors. for example, ellipse deadzone should not have any axis snapping but still do range shifting by rescaling input magnitude to zero at the edge of the deadzone.

## Related work

Lots of other deadzones that can be supported in the future: https://imgur.com/a/xSfcP (most interesting concepts are Back 4 Blood's unregistered diagonal wedges and Anarchy Reigns' two-threshold central deadzone where once you start moving, a smaller threshold must be returned to to stop moving)

100-TomatoJuice:
Closes #398

The issue above describes the changes made and why really well.

To-Do:
- [x] Fix tests
- [x] Finish updating docs
- [x] Turn back on `missing_docs`
- [x] Update `RELEASES.md`

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -8,6 +8,13 @@
 - Added `InputMap::Clear`.
 - Fixed [a bug](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/430) related to incorrect axis data in `Chord` when not all buttons are pressed.
 
+### Enhancements
+
+- Changed `DualAxis` and `SingleAxis` deadzones to be more accessible. 
+  - All deadzones now scale the input so that it is continuous.
+  - `DeadZoneShape::Cross` handles each axis seperately, making a per-axis "snapping" effect.
+  - An input on the line of a deadzone is now considering inside.
+
 ## Version 0.11.2
 
 - fixed [a bug](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/285) with mouse motion and mouse wheel events being improperly counted
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -10,7 +10,7 @@
 
 ### Enhancements
 
-- Changed `DualAxis` and `SingleAxis` deadzones to be more accessible. 
+- Improved deadzone handling for both `DualAxis` and `SingleAxis` deadzones 
   - All deadzones now scale the input so that it is continuous.
   - `DeadZoneShape::Cross` handles each axis seperately, making a per-axis "snapping" effect.
   - An input that falls on the exact boundary of a deadzone is now considered inside it.
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -13,7 +13,8 @@
 - Changed `DualAxis` and `SingleAxis` deadzones to be more accessible. 
   - All deadzones now scale the input so that it is continuous.
   - `DeadZoneShape::Cross` handles each axis seperately, making a per-axis "snapping" effect.
-  - An input on the line of a deadzone is now considering inside.
+  - An input that falls on the exact boundary of a deadzone is now considered inside it.
+
 
 ## Version 0.11.2
 
--- a/src/axislike.rs
+++ b/src/axislike.rs
@@ -743,11 +743,11 @@ pub enum DeadZoneShape {
     /// The cross is represented by two rectangles.
     Cross {
         /// The width of the horizonal axis.
-        /// 
+        ///
         /// Affects the snapping of the y-axis.
         horizontal_width: f32,
         /// The width of the vertical axis.
-        /// 
+        ///
         /// Affects the snapping of the x-axis.
         vertical_width: f32,
     },
@@ -809,7 +809,8 @@ impl DeadZoneShape {
         if new_value == Vec2::ZERO {
             return None;
         } else {
-            let scaled_value = Self::scale_value(new_value, Vec2::new(vertical_width, horizontal_width));
+            let scaled_value =
+                Self::scale_value(new_value, Vec2::new(vertical_width, horizontal_width));
             return Some(DualAxisData::from_xy(scaled_value));
         }
     }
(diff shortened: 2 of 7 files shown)


src/axislike.rs
```diff
@@ -797,53 +781,58 @@ impl std::hash::Hash for DeadZoneShape {
...
+    }
+
+    /// Computes the input value based on the ellipse deadzone.
+    fn ellipse_deadzone_value(
+        &self,
+        value: Vec2,
+        radius_x: f32,
+        radius_y: f32,
+    ) -> Option<DualAxisData> {
+        let radius_x = radius_x.max(0.0001);
```
100-TomatoJuice, line 825: I just picked a random small number to make sure that we aren't dividing by 0.0. Is there is more standard number that is typically used in this situation, like epsilon? I don't think I can return `None` or `Some` when the deadzone size is 0.0 as then all values are ignored or read.

alice-i-cecile: f32::EPSILON is probably the most "standard", but just extracting it into a constant called `EPSILON` is probably clear enough.

src/axislike.rs
```diff
@@ -732,32 +740,28 @@ impl From<DualAxisData> for Vec2 {
...
-    /// Deadzone with the shape of a rectangle.
-    Rect {
-        /// The width of the rectangle.
-        width: f32,
-        /// The height of the rectangle.
-        height: f32,
+        /// The width of the horizonal axis.
+        ///
+        /// Affects the snapping of the y-axis.
+        horizontal_width: f32,
```
100-TomatoJuice, line 760: There can probably be a better name to describe the settings for the axes?

100-TomatoJuice:
@rodolphito, if you do not mind/have to the time, a review from you would be helpful to ensure that this pr is consistent with your issue. :)

RELEASES.md
```diff
@@ -8,6 +8,13 @@
 - Added `InputMap::Clear`.
 - Fixed [a bug](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/430) related to incorrect axis data in `Chord` when not all buttons are pressed.
 
+### Enhancements
+
+- Changed `DualAxis` and `SingleAxis` deadzones to be more accessible.
```
alice-i-cecile, line 13: "More accessible" is really unclear to the reader.

src/input_streams.rs
```diff
@@ -266,14 +268,24 @@ impl<'a> InputStreams<'a> {
...
         // Helper that takes the value returned by an axis and returns 0.0 if it is not within the
         // triggering range.
-        let value_in_axis_range = |axis: &SingleAxis, value: f32| -> f32 {
-            if value >= axis.negative_low && value <= axis.positive_low && include_deadzone {
-                0.0
-            } else if axis.inverted {
-                -value * axis.sensitivity
-            } else {
-                value * axis.sensitivity
+        let value_in_axis_range = |axis: &SingleAxis, mut value: f32| -> f32 {
```
100-TomatoJuice, line 271: This is a little gross but otherwise the single axis scaling would affect the dual axis scaling. Perhaps there is a better way I did not think about?

RELEASES.md
100-TomatoJuice: That's true. I could remove the that bullet point and just list the points underneath? I hoping that the following bullets would explain it, but I understand how that can be unclear.

alice-i-cecile:
```suggestion
- Improved deadzone handling for both `DualAxis` and `SingleAxis` deadzones 
```

RELEASES.md
```diff
@@ -8,6 +8,13 @@
 - Added `InputMap::Clear`.
 - Fixed [a bug](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/430) related to incorrect axis data in `Chord` when not all buttons are pressed.
 
+### Enhancements
+
+- Changed `DualAxis` and `SingleAxis` deadzones to be more accessible. 
+  - All deadzones now scale the input so that it is continuous.
+  - `DeadZoneShape::Cross` handles each axis seperately, making a per-axis "snapping" effect.
+  - An input on the line of a deadzone is now considering inside.
```
alice-i-cecile, line 16:
```suggestion
  - An input on the boundary of a deadzone is now considering inside.

```

PR #442, opened 2024-01-10
Fix EguiUserTextures related crash when gui feature is nabled and bevy-egui is used

mrafaj:
When the egui feature is enabled and the egui flag is used on a project using both bevy-egui and leafwing-input-manager, the program crashes with the following error:

```rust
thread 'Compute Task Pool (3)' panicked at C:\Users\Korisnik\.cargo\registry\src\index.crates.io-6f17d22bba15001f\bevy_ecs-0.12.1\src\system\system_param.rs:544:17:
Resource requested by leafwing_input_manager::systems::update_action_state<egui_test::Action> does not exist: bevy_egui::EguiUserTextures
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `leafwing_input_manager::systems::update_action_state<egui_test::Action>`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
```

This fixes the crash by replacing the request for EguiContexts done by leafwing_input_manager::systems::update_action_state with a query for EguiContext, as EguiContexts is jut a shortcut to getting to both the EguiUserTextures resource and the EguiContext query with maybe_egui.ctx_mut() returning the Context object of Egui's primary window. This PR fixes the crash as the EguiUserTextures resource is no longer being requested since it doesn't need to be for the egui feature.

### DIFF
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -29,7 +29,7 @@ use crate::action_state::ActionDiffEvent;
 #[cfg(feature = "ui")]
 use bevy::ui::Interaction;
 #[cfg(feature = "egui")]
-use bevy_egui::EguiContexts;
+use bevy_egui::EguiContext;
 
 /// Advances actions timer.
 ///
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
Great! I've seen reports of this but couldn't track it down.

PR #449, opened 2024-01-23
Use (Action, Input) ordering for `InputMap` construction

alice-i-cecile:


src/input_map.rs
```diff
@@ -514,7 +482,7 @@ impl<A: Actionlike> From<HashMap<A, Vec<UserInput>>> for InputMap<A> {
 impl<A: Actionlike> FromIterator<(A, UserInput)> for InputMap<A> {
     /// Create `InputMap<A>` from iterator with item type `(A, UserInput)`
     fn from_iter<T: IntoIterator<Item = (A, UserInput)>>(iter: T) -> Self {
-        InputMap::new(iter.into_iter().map(|(action, input)| (input, action)))
+        InputMap::new(iter.into_iter())
```
alice-i-cecile, line 485:
```suggestion
        InputMap::new(iter)
```

PR #454, opened 2024-01-28
Restore the `ButtonState::Pressed` state on re-enabling `ToggleActions`

### LINKED ISSUE #446: Held inputs become `JustPressed` when `ToggleActions` is enabled
## Version

`leafwing-input-manager` 0.11, bevy `0.12`.

## Operating system & version

Arch Linux

## What you did

Setup: `KeyCode::R` is mapped to `GameAction::Restart`, with a system that restarts the game on `action_just_pressed(GameAction::Restart)`. Additionally, there is a system that disables `GameAction` on `KeyCode::Slash` and re-enables it on `KeyCode::Return`.

I press `/` to disable `GameAction`, then press and hold `R`, then press `Return` while `R` is still held.

When `GameAction` is re-enabled, the action state for `GameAction::Restart` should go from `Released` to `Pressed` -- because `R` is still being held from an earlier press -- thus not triggering a game restart.

## What actually happened

The game action went to `JustPressed`, triggering a game restart.

Shute052:
# Objective

- Fixes #446, caused by the lack of restoration for `ButtonState::Pressed` state when re-enabling `ToggleActions`.

## Solution

- Add a new state, `ButtonState::PressedWhenDisabling`, to retain the `Pressed` state for subsequent restoration when `ToggleActions` is re-enabled.
- Add a new state, `ButtonState::Disabled`, to mark the action is disabled by `ToggleActions`.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -22,6 +22,7 @@
 - registered types in the reflection system
 - added `InputMap::clear`
 - fixed [a bug](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/430) related to incorrect axis data in `Chord` when not all buttons are pressed.
+- fixed [a bug](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/446) related to missing restoration for `ButtonState::Pressed` state when re-enabling `ToggleActions`.
 
 ### Code Quality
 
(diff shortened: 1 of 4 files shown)


benfrankel:
I believe this doesn't handle the case where the input is _not_ pressed when disabling, and is instead pressed and held while `ToggleActions` is already disabled.

Shute052:
> I believe this doesn't handle the case where the input is _not_ pressed when disabling, and is instead pressed and held while `ToggleActions` is already disabled.

It appears to be fixed by the new commit on my end, but the solution has become less tidy.

PR #455, opened 2024-01-29
Replace unnecessary `Vec` with a simple flag in `InputMap::which_pressed()`

Shute052:
# Objective

The `InputMap::which_pressed()` function is invoked by the `update_action_state()` system every tick when the corresponding `Actionlike` is enabled. Currently, for press state checking, it unnecessarily accumulates a temporary vector, causing avoidable performance losses.

## Solution

Replace the temporary `inputs` vector with a simple `pressed` flag to enhance the efficiency of press state checking.

### DIFF
--- a/src/input_map.rs
+++ b/src/input_map.rs
@@ -323,20 +323,20 @@ impl<A: Actionlike> InputMap<A> {
 
         // Generate the raw action presses
         for (action, input_vec) in self.iter() {
-            let mut inputs = Vec::new();
+            let mut pressed = false;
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
Yep, this looks good :) I think that this was a relic from when I was trying to track reasons_pressed.

alice-i-cecile:
Is there anything else you'd like to do before I merge this in?

Shute052:
> Is there anything else you'd like to do before I merge this in?

Nope, perhaps no further actions

Issue #416, opened 2023-11-16
ToggleActions is a resource, even with InputMap as component


NiseVoid:
## Which feature is frustrating to use or confusing?

Using ToggleActions feels unnatural to use, and could lead to problems in some couch co-op scenarios, with component-based input maps, because they are all resources that apply globally.

## Expectation

ToggleActions as components, or maybe a way to toggle the input maps directly?

Veritius:
I think that `ToggleActions<T>` should be both a resource *and* a component. This lets you implement a 'paused' state by adding the resource, affecting all entities, but also letting you enable/disable it per-entity.

alice-i-cecile:
Totally agree.

benfrankel:
It seems like this would be fixed by https://github.com/Leafwing-Studios/leafwing-input-manager/pull/456.

PR #456, opened 2024-01-29
Replace `ToggleActions<A>` with per-action disable flags

benfrankel:
# Objective

Fixes https://github.com/Leafwing-Studios/leafwing-input-manager/issues/446, https://github.com/Leafwing-Studios/leafwing-input-manager/issues/416.

# Solution

Don't disable the systems that update action state. Instead, include a `disabled` flag per action within `ActionState<A>`. While `disabled` is `true`, an action will always read as released, regardless of its actual state.

Pros:
- Replaces `ToggleActions<A>`, making the API surface area smaller.
- Supports granular enabling / disabling between:
  - The individual actions of an `ActionState<A>`
  - Different entities with the same `ActionState<A>` component type
  - Resource vs component instances of `ActionState<A>`
- Keeps the action update logic simple and consistent by only affecting how the state is read, not how it's written.
- Similar to the existing `consumed` flag.

Cons:
- Harder to enable / disable _all actions_ for a given `A: Actionlike` (you'd have to check for a resource as well as query for any components -- and that still doesn't account for e.g. new entities spawning with their own `ActionState<A>` in later frames).
- Disabling an action only affects its button state (pressed, released, etc.), not e.g. `value` and `axis_pair` (this is also the current behavior without this PR, see https://github.com/Leafwing-Studios/leafwing-input-manager/issues/506). Those could be updated as well to read as zero while disabled.
- Breaking change.

Alternative approach: https://github.com/Leafwing-Studios/leafwing-input-manager/pull/454.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -7,7 +7,7 @@
 - the `timing` field of the `ActionData` is now disabled by default. Timing information will only be collected
 if the `timing` feature is enabled. It is disabled by default because most games don't require timing information.
 (how long a button was pressed for)
-
+- removed `ToggleActions` resource in favor of new methods on `ActionState`: `disable_all`, `disable(action)`, `enable_all`, `enable(action)`, and `disabled(action)`.
 - removed `InputMap::build` method in favor of new fluent builder pattern (see 'Usability: InputMap' for details).
 - renamed `InputMap::which_pressed` method to `process_actions` to better reflect its current functionality for clarity.
 - replaced axis-like input handling with new input processors (see 'Enhancements: Input Processors' for details).
@@ -96,6 +96,7 @@ Input processors allow you to create custom logic for axis-like input manipulati
 
 ### Bugs
 
+- fixed a bug where enabling a pressed action would read as `just_pressed`, and disabling a pressed action would read as `just_released`.
 - fixed a bug in `InputStreams::button_pressed()` where unrelated gamepads were not filtered out when an `associated_gamepad` is defined.
 
 ## Version 0.13.3
--- a/benches/action_state.rs
+++ b/benches/action_state.rs
@@ -69,6 +69,7 @@ fn criterion_benchmark(c: &mut Criterion) {
                     #[cfg(feature = "timing")]
                     timing: Timing::default(),
                     consumed: false,
+                    disabled: false,
                 },
             )
         })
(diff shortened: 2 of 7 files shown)


Shute052:
It seems more suitable for me because my solution is just to keep the public API from making significant changes.

Shute052:
Would it be better to add a `disabled` flag and a `toggle_all()` function within the `ActionState` to ensure similarity with the existing public API?

benfrankel:
> Would it be better to add a `disabled` flag and a `toggle_all()` function within the `ActionState` to ensure similarity with the existing public API?

In addition to the per-action flags, or instead of them? I was originally going to only put one `disabled` flag in `ActionState` itself, but while implementing that I figured per-action would be a little bit easier to implement (because there are multiple places that access `action_data` directly) and would give the user more control.

Shute052:
Ahh, the newly added `enabled()` and `disabled()` meets my needs, which weren't there before.

benfrankel:
Rebased and fixed merge conflicts.

atlv24:
This doesn't trigger just_pressed/just_released when enabling/disabling a currently-pressed button, does it?

benfrankel:
> This doesn't trigger just_pressed/just_released when enabling/disabling a currently-pressed button, does it?

Nope, while an action is disabled, it will _strictly_ read as `released` and nothing else.

atlv24:
That might be undesirable, whats the justification? I would expect disabling a pressed button to release it. This would allow multiple `just_pressed` frames to be observed without any `just_released` frames in-between, or vice-versa

benfrankel:
The reasoning is elaborated on in https://github.com/Leafwing-Studios/leafwing-input-manager/issues/446 and https://github.com/Leafwing-Studios/leafwing-input-manager/issues/416.

It may seem wrong but the alternative (the current behavior) is also wrong. An action wasn't actually "just pressed" just because you re-enabled it, and you don't want re-enabling an action to trigger all of its `do_thing.run_if(action_just_pressed)` systems.

atlv24 (approved):

alice-i-cecile (changes requested):
This needs release notes before it can be merged. I'm in favor of this design though, so I'll write it myself in the preparation for the release if you don't have a chance.

benfrankel:
Wrote release notes.

alice-i-cecile (approved):

PR #457, opened 2024-01-30
Fix incorrect `ActionData::axis_pair` value when handling multiple `DualAxisData`

Shute052:
# Objective

The previous PR (#450) removed the accumulation for `ActionData::axis_pair` in `InputMap::which_pressed()`, resulting in a problem where multiple `DualAxisData` updates would only preserve the last one due to the loss of previously recorded `DualAxisData`.

## A Simple Reproduction Case

```rust
InputMap::new([
        (Action, VirtualDPad::arrow_keys()),
        (Action, VirtualDPad::wasd()),
]);
```

While it functions correctly when pressing WASD, if only the arrow keys were pressed, `ActionState::axis_pair()` will return `None`.

## Solution

Re-implement the accumulation of `ActionData::axis_pair` in `InputMap::which_pressed()`.

## Changelog

- Re-add the accumulation of `ActionData::axis_pair`
- Simplify the `ButtonState::JustPressed` check by moving it into the `input_pressed` checking block and remove the `pressed` flag

### DIFF
--- a/src/input_map.rs
+++ b/src/input_map.rs
@@ -326,9 +326,13 @@ impl<A: Actionlike> InputMap<A> {
             let mut action_datum = ActionData::default();
 
             for input in input_vec {
-                if let Some(new_axis_pair) = input_streams.input_axis_pair(input) {
-                    let current_axis_pair = action_datum.axis_pair.unwrap_or_default();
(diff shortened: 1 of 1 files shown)


Hellzbellz123:
this fixes the issue i commented about on discord, glad someone figured out what was actually wrong

alice-i-cecile:
Thank you!

PR #466, opened 2024-02-19
Refactor for clearer and more performant code

Shute052:
Simplified PR #463 to refactor for clearer and more performant code.

### Logic Refactor

- Refactored `DeadZoneShape::cross_deadzone_value()` and `DeadZoneShape::ellipse_deadzone_value()` to improve clarity of logic.
- Refactored `InputStreams::button_pressed()` and `InputStreams::input_value()` functions for both `MouseWheel` and `MouseMotion` events by using more efficient `f64` accumulation, reducing unnecessary `match`.

### Deduplication

- Extracted `ActionState::action_data_mut_or_default()` function to remove duplication in the parts checking `press()`, `release()`, and `consume()`.
- Extracted `deadzone_axis_value()` to remove duplication in the parts applying deadzone to input values.
- Reused the existing `virtual_axis_button_clash()` for checking clashes of `VirtualAxis`.
- Extracted `send_keyboard_input()`, `send_mouse_wheel()`, `send_mouse_motion()`, `send_gamepad_button_changed()` for `MutableInputStreams` when mocking input.
- Extracted `collect_events_cloned()` to remove duplication in the parts collecting the cloned events of `MouseWheel` and `MouseMotion`.
- Reused the existing `InputStreams::extract_dual_axis_data()` to check `InputKind::DualAxis` has been `button_pressed()`.
- Extracted `get_gamepad_value()` closure to reuse the logic getting values from `InputStreams::gamepad_axes` and `InputStreams::gamepad_buttons_axes`.
- Extracted `InputStreams::extract_single_axis_data()` for computing the `input_value()` of `VirtualAxis` and the `input_axis_pair()` of `VirtualDpad`.
- Extracted `RawInputs::merge_input_data()` to remove duplication in the parts merging `UserInput`s into a `RawInputs`.

### Code Quality

- Used the existing std::consts like `FRAC_PI_2` rather than magic numbers.
- Used `f32::from()` rather than manual `if` when just getting a `f32` from `bool` value.
- Used `matches!` rather than manual `match` when just getting a `bool` field value.
- Used `if let Some()` rather than manual `match` when the `None` arm does nothing.
- Used `Result::ok()` rather than manual `match` when casting `Result` into `Option`.
- Used `Option::then()` rather than manual `match` when just mapping its value.
- Used `bool::then_some()` rather than `if` when just mapping it into `Option`.
- Used iterators rather than manual `for` loops when the logic can be replaced with simple and meaningful operators.
[...]

PR #474, opened 2024-02-19
Prefer `HashMap::values_mut()`

Shute052:
Continue with PR #466.


Shute052:
If there's nothing else to change, let's get ready for 0.12.1.

PR #478, opened 2024-02-20
Runtime disabling of specific input type tracking

Shute052:
## Objective

- Parts of completing #483

Allow disabling of specific input type tracking.
This can be especially useful if users encounter conflicts with input handling from other crates not currently integrated yet.

### Solution

Add `TrackingState` enum for always enabling, always disabling, or ignoring once for input type tracking.

Add a new resource called `TrackingInputType`, which includes flags for enabling specific input type tracking.

Furthermore, add run conditions that are active when the `TrackingInputType` is enabled for specific input types.

### Changelog

- Added `TrackingState` enum 
- Added `TrackingInputType` Resource
- Exported new items into the `prelude` module
- Extracted `prioritize_ui_input` and `prioritize_egui_input` systems from `update_action_state`
- Integrated new systems into `InputManagerPlugin`
- `InputStreams` now take `Option` for gamepad values

### DIFF
--- a/src/plugin.rs
+++ b/src/plugin.rs
@@ -8,6 +8,10 @@ use crate::axislike::{
 use crate::buttonlike::{MouseMotionDirection, MouseWheelDirection};
 use crate::clashing_inputs::ClashStrategy;
 #[cfg(feature = "egui")]
+use crate::common_conditions::tracking_keyboard_input;
+#[cfg(any(all(feature = "ui", not(feature = "no_ui_priority")), feature = "egui"))]
+use crate::common_conditions::tracking_mouse_input;
+#[cfg(feature = "egui")]
 use crate::conflicting_inputs::prioritize_egui_inputs;
 #[cfg(all(feature = "ui", not(feature = "no_ui_priority")))]
 use crate::conflicting_inputs::prioritize_ui_inputs;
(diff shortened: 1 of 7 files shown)


Shute052:
@alice-i-cecile This isn't related to #185, but could you take a look at the parts about `configure_sets`? I'm not sure if it's going to work like I expect.

src/plugin.rs
```diff
@@ -225,6 +257,11 @@ impl<A: Actionlike> Default for ToggleActions<A> {
 /// `Reset` must occur before `Update`
 #[derive(SystemSet, Clone, Hash, Debug, PartialEq, Eq)]
 pub enum InputManagerSystem {
+    /// Performs actions before [`InputManagerSystem::Update`]
+    ///
+    /// This is useful for temporarily disabling [`TrackingInputType`],
+    /// and it will be re-enabled after handling all conflicting inputs
+    PreUpdate,
```
alice-i-cecile, line 264: This is a confusing name: it conflicts with the `Schedule`.

alice-i-cecile:
Okay, I definitely think this is a useful tool, and I really like the notion of having a way to keep track of which input streams we're subscribed to.

However, I don't want to add more complexity/features to the code until #321 is solved. I think that this is helpful to shape that sort of design though.

src/plugin.rs
Shute052: Remove this by adding `TrackingState` enum

Shute052:
Blocked until the new architecture is implemented

alice-i-cecile:
Closing: this will be easier to redo.

PR #479, opened 2024-02-20
Export `action_data_mut_or_default`

Shute052:
Export `action_data_mut_or_default`, which behaves the same way with `action_data_mut` in LWIM 0.11.
Just another unmentioned breaking changes of 0.12.

This should be particularly helpful for those working with setups similar to [this](https://discord.com/channels/691052431525675048/1034547742262951966/1209616393402581063).

### DIFF
--- a/src/action_state.rs
+++ b/src/action_state.rs
@@ -207,6 +207,20 @@ impl<A: Actionlike> ActionState<A> {
         self.action_data.get_mut(action)
     }
 
+    /// A mutable reference of the [`ActionData`] of the corresponding `action`.
+    ///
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

Shute052:
Or change the existing code to reflect its real behavior?
- `action_data` to `action_data_if_populated`
- `action_data_mut` to `action_data_mut_if_populated`
- `action_data_mut_or_default` to `action_data_mut`

alice-i-cecile:
Let's keep this as is for now so we can ship this without breaking changes.

Issue #443, opened 2024-01-17
ActionDiffs are not being networking correctly when an action is consumed


cBournhonesque:
## Version

[The release number or commit hash of the version you're using.](https://github.com/Leafwing-Studios/leafwing-input-manager/commit/cda8d12b2c25ec39b3fb57911d5b9591eadb9224)

## What you did

In my `Update` schedule, I have something like:
```
if action.pressed(PlayerActions::Shoot) {
   action.consume();
}
```
and in `PreUpdate`, I generate `ActionDiffs` using the diff-generation system after `InputSystemSet::Update`.

However I don't get any diffs for the action that was consumed.

My understanding is that the flow is this:
- Frame 1: user presses button
  - PreUpdate: PressButton, Shoot is `pressed`
  - Update: Shoot is `consumed` (`consumed=true` and `state=JustReleased`)
- Frame 2
  - PreUpdate: Tick, so we go from `JustReleased` to `Released`
  - PreUpdate: in generate-diffs, we only network JustReleases, so this is not networked
...
- Frame X: user releases button
  - PreUpdate: `consumed=false`. `state=Released` and not `JustReleased` because the button was already marked as `Released`. The diff-generation system only networks `JustReleased` events, so I never get the information that the key was released.

## What you expected to happen

The release of the button is networked, either when the button is consumed, or when I actually released the key.

## What actually happened

Nothing is networked.

## Potential solutions:

- I could try to be more clever in where I use the diff-generation system (maybe run it once before `InputSystem::Tick` as well so that the consumed action is still JustReleased

- Maybe when a key was consumed, we still trigger a `JustReleased` when the actual key is `Released`.
(so `release()` would be something like:
```
if *self != ButtonState::Released || self.consumed {
            *self = ButtonState::JustReleased;
}
```

- maybe in the diff generation system, we send a `release` diff for every action that is consumed, even if it's not a `JustRelease`. We could get potentially a lot of duplicate `Release` `ActionDiffs`, but it's a good short-term solution

Shute052:
isn't this fixed by #480?

cBournhonesque:
Nope; the problem is not with the `consume` function, it's more about how ActionDiffs are generated.

the problem is that consumed actions are immediately marked as JustReleased, and then become `Released` in the next frame (because we tick them once).
The generate_action_diffs runs after that and doesn't generate diff for them, since it only generates diffs for `JustReleased` and `JustPressed` actions.

[...]

PR #480, opened 2024-02-21
Add missing handling for `ActionState::consumed`

Shute052:
Missing handling with `ActionState::consumed` was added back, which was overlooked in #450.

image

To be specific, `ActionState::press(Action)` takes `consumed` into account, whereas `ActionData::state::press(Action)` did not.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -8,6 +8,12 @@
   - by default, this library will prioritize `bevy::ui`.
   - if you want to disable this priority, add the newly added `no_ui_priority` feature to your configuration.
 
+## Version 0.13.1
+
+### Bugs
+
+- fixed a bug related to missing handling for `ActionState::consumed`
+
 ## Version 0.13.0
 
 ### Breaking Changes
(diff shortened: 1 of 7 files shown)


Shute052:
Both 0.12.2 and 0.13.1?

RELEASES.md
```diff
@@ -27,6 +27,12 @@
 - `bevy` dependency has been bumped from 0.12 to 0.13.
 - `bevy_egui` dependency has been bumped from 0.24 to 0.25.
 
+## Version 0.12.2
```
alice-i-cecile, line 30: This should probably be `0.13.1`. I don't think this is impactful enough to backport without others asking for it.

alice-i-cecile (changes requested):
Change itself looks good, but let's aim for 0.13.1 instead.

Shute052:
Also bump the version number

alice-i-cecile (approved):

Shute052:
This fixed #443

Issue #483, opened 2024-02-22
Refactoring underlying architecture to support various input types


Shute052:
# Motivation

Setting the stage for future additions (maybe):

- cursor position related to the game window and game world (like example mouse_position.rs and twin_stick_controller.rs)
- cursor edge panning
- #223
- touchscreen capabilities (existing in Bevy)
- #224
- cancellable actions with fallthrough
- #485
- #372
- #505
- XR inputs
- adding support for other powerful plugins such as #15

This also gives users the freedom to tailor the input kinds they want.

# Tasks

- [x] #513
- [x] #515
- [ ] #490
- [ ] #456
- [ ] #478
- [ ] #185

# Undecided Changes

## `InputStream` trait

Replace the large `InputStreams` with smaller, single-responsibility implementations:

```rust
/// Allows defining a specific kind of stream collecting user inputs.
pub trait InputStream<'a> {
    /// Collects the inputs from a [`World`] for testing.
    fn from_world(world: &'a World) -> Self;
}

/// Bulit-in support for `ButtonInput<KeyCode>`.
pub type KeyCodeStream = ButtonInput<KeyCode>
impl<'a> InputStream<'a> for ButtonInput<KeyCode> {
    fn from_world(world: &'a World) -> Self {
        world.resource::<Self>().clone()
    }
}
```

## `InputQuery`

Similar to Bevy's `Query`:

```rust
fn button_pressed(input_query: InputQuery<KeyCodeStream>, button: KeyCode) -> bool {
    let keycode_input: KeyCodeStream = input_query.get();
    keycode_input.pressed(button)
}
```

Shute052:
It's a huge task, but I'm hoping this setup will work out. This makes it easy to add new stuff later on, like touchscreens and gestures, and allowing users to customize their input types.

 Any other ideas for now?

Shute052:
Should we prioritize removing `input_mocking.rs` first? See [the comment](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/321#issuecomment-1548827763).

alice-i-cecile:
Not fully sold on the rename: I like the way that the current API makes the state machine nature clear.

Otherwise this looks good. And yes, we should swap out for a lighter solution to input mocking for tests: I think that will make the refactoring much easier.

alice-i-cecile:
> And how do I make all the references look like the one above? Even though the content in markdown is the same!

[...]

PR #490, opened 2024-02-23
Refactor `UserInput` and `InputKind` variants into separate implementations

### LINKED ISSUE #321: Represent user inputs via traits, rather than a large enum

Shute052:
## Objective

- Parts of completing #483
- Closes #321
- Continues with #346

The current `UserInput` and `InputKind` enums are overly large and challenging to extend with new input types.

## Solution

Introduce a new trait called `UserInput` trait.

Refactor the `UserInput` and `InputKind` variants into separate impls of the new `UserInput` trait, each focusing on a single responsibility.

## TODO

- [x] Fix broken clashing check.
- [x] Documentation and doc examples.

## Breacking Changes

- renamed `Modifier` enum to `ModifierKey`, avoiding potential naming conflicts with other libraries or parts of your codebase.
- refactored `InputKind` variants for representing user inputs, it now represents the kind of data an input can provide (e.g., button-like input, axis-like input).
- removed `UserInput` enum in favor of the new `UserInput` trait and its impls.
- by default, all inputs are unprocessed now, using `With*ProcessingPipelineExt` methods to define your preferred processing steps.
- applied clashing check to continuous inputs, for example:
  - `MouseScrollAxis::Y` will clash with `MouseScrollDirection::UP` and `MouseScrollDirection::DOWN`.
  - `MouseMove` will clash with all individual axes and all directional events.

## Changelog

- added `UserInput` trait.
- added `UserInput` impls for gamepad input events:
  - implemented `UserInput` for Bevy’s `GamepadAxisType`-related inputs.
    - `GamepadStick`: Continuous or discrete movement events of the left or right gamepad stick along both X and Y axes.
    - `GamepadControlAxis`: Continuous or discrete movement events of a `GamepadAxisType`.
    - `GamepadControlDirection`: Discrete movement direction events of a `GamepadAxisType`, treated as a button press.
  - implemented `UserInput` for Bevy’s `GamepadButtonType` directly.
  - added `GamepadVirtualAxis`, similar to the old `UserInput::VirtualAxis` using two `GamepadButtonType`s.
  - added `GamepadVirtualDPad`, similar to the old `UserInput::VirtualDPad` using four `GamepadButtonType`s.
- added `UserInput` impls for keyboard inputs:
  - implemented `UserInput` for Bevy’s `KeyCode` directly.
  - implemented `UserInput` for `ModifierKey`.
  - added `KeyboardVirtualAxis`, similar to the old `UserInput::VirtualAxis` using two `KeyCode`s.
  - added `KeyboardVirtualDPad`, similar to the old `UserInput::VirtualDPad` using four `KeyCode`s.
- added `UserInput` impls for mouse inputs:
  - implemented `UserInput` for movement-related inputs.
    - `MouseMove`: Continuous or discrete movement events of the mouse both X and Y axes.
    - `MouseMoveAxis`: Continuous or discrete movement events of the mouse on an axis, similar to the old `MouseMotionAxisType`.
    - `MouseMoveDirection`: Discrete movement direction events of the mouse on an axis, similar to the old `MouseMotionDirection`.
  - implemented `UserInput` for wheel-related inputs.
    - `MouseScroll`: Continuous or discrete movement events of the mouse wheel both X and Y axes.
    - `MouseScrollAxis`: Continuous or discrete movement events of the mouse wheel on an axis, similar to the old `MouseWheelAxisType`.
    - `MouseScrollDirection`: Discrete movement direction events of the mouse wheel on an axis, similar to the old `MouseWheelDirection`.
- added `InputChord` for combining multiple inputs, similar to the old `UserInput::Chord`. 

## Migration Guide

- The old `SingleAxis` is now:
  - `GamepadControlAxis` for gamepad axes. 
  - `MouseMoveAxis::X` and `MouseMoveAxis::Y` for continuous mouse movement.
  - `MouseScrollAxis::X` and `MouseScrollAxis::Y` for continuous mouse wheel movement.
- The old `DualAxis` is now:
  - `GamepadStick` for gamepad sticks.
  - `MouseMove::default()` for continuous mouse movement.
  - `MouseScroll::default()` for continuous mouse wheel movement.
- The old `Modifier` is now `ModifierKey`.
- The old `MouseMotionDirection` is now `MouseMoveDirection`.
- The old `MouseWheelDirection` is now `MouseScrollDirection`.
- The old `UserInput::Chord` is now `InputChord`.
- The old `UserInput::VirtualAxis` is now:
  - `GamepadVirtualAxis` for four gamepad buttons.
  - `KeyboardVirtualAxis` for four keys.
  - `MouseMoveAxis::X.digital()` and `MouseMoveAxis::Y.digital()` for discrete mouse movement.
  - `MouseScrollAxis::X.digital()` and `MouseScrollAxis::Y.digital()` for discrete mouse wheel movement.
- The old `UserInput::VirtualDPad` is now:
  - `GamepadVirtualDPad` for four gamepad buttons.
  - `KeyboardVirtualDPad` for four keys.
  - `MouseMove::default().digital()` for discrete mouse movement.
  - `MouseScroll::default().digital()` for discrete mouse wheel movement.

## Next Steps

- Refactor `InputStreams` into separate implementations.
- Find alternative approaches for input mocking that provide more flexibility for future input type extensions, replacing `RawInputs`


Shute052:
If there are any parts of the document where examples should be added, please let me know directly.
Because I am the author of the document, I'm not sure where examples should be added to ensure clarity for the readers.
However, adding examples to every part would be too much work, and my vacation is ending soon.
I can only do as much as I can before then

Shute052:
- This PR is complete and awaits the merging of #521.

src/user_input/gamepad.rs
```diff
@@ -0,0 +1,1150 @@
...
+/// - Gamepad Selection: By default, reads from **any connected gamepad**.
+///     Use the [`InputMap::set_gamepad`] for specific ones.
+/// - Activation: Only if the axis is currently held in the chosen direction.
+/// - Single-Axis Value:
+///   - `1.0`: The input is currently active.
+///   - `0.0`: The input is inactive.
+///
+/// [`InputMap::set_gamepad`]: crate::input_map::InputMap::set_gamepad
+///
+/// ```rust,ignore
```
Shute052, line 54: Doc tests for this module are ignored, due to the excessive code required to add usable handles for each example.
I feel that they should be focusing on describing input behavior

alice-i-cecile:
@Shute052 I'd like to review and merge this this week: please resolve merge conflicts if you get a chance and let me know if there's anything else that remains to be done.

src/clashing_inputs.rs
```diff
@@ -47,36 +46,87 @@ impl ClashStrategy {
...
-        use UserInput::*;
+/// The basic inputs that make up a [`UserInput`].
+#[derive(Debug, Clone)]
+#[must_use]
+pub enum BasicInputs {
+    /// The input consists of a single, fundamental [`UserInput`].
+    /// In most cases, the input simply holds itself.
+    Simple(Box<dyn UserInput>),
+
+    /// The input consists of multiple independent [`UserInput`]s.
```
alice-i-cecile, line 57: I don't understand the distinction between Composite and Group.

src/clashing_inputs.rs
```diff
@@ -47,36 +46,87 @@ impl ClashStrategy {
...
+        }
+    }
+
+    /// Returns the number of the logical [`UserInput`]s that make up the input.
+    #[allow(clippy::len_without_is_empty)]
+    #[inline]
+    pub fn len(&self) -> usize {
+        match self {
+            Self::Simple(_) => 1,
+            Self::Composite(_) => 1,
```
alice-i-cecile, line 81: It's quite surprising that this is 1. If it's not a bug, it needs a doc comment.

src/input_map.rs
```diff
@@ -200,51 +215,22 @@ impl<A: Actionlike> InputMap<A> {
...
     }
 
-    /// Insert a mapping between the simultaneous combination of `buttons` and the `action` provided
-    ///
-    /// Any iterator convertible into a [`InputKind`] can be supplied,
-    /// but will be converted into a [`HashSet`](bevy::utils::HashSet) for storage and use.
-    /// Chords can also be added with the [insert](Self::insert) method if the [`UserInput::Chord`] variant is constructed explicitly.
-    ///
-    /// When working with keyboard modifier keys, consider using the `insert_modified` method instead.
-    pub fn insert_chord(
```
alice-i-cecile, line 226: It would be nice to keep these two convenience methods if possible.

src/raw_inputs.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use crate::axislike::DualAxisType;
+use crate::user_input::{GamepadControlDirection, MouseMoveDirection, MouseScrollDirection};
+
+/// The basic input events that make up a [`UserInput`](crate::user_input::UserInput).
+///
+/// Typically obtained by calling [`UserInput::raw_inputs`](crate::user_input::UserInput::raw_inputs).
+#[derive(Default, Debug, Clone, PartialEq)]
+#[must_use]
+pub struct RawInputs {
```
alice-i-cecile, line 16: Hmm. I don't love this design. Doesn't this just get us back to the same place, where `UserInput` can't be reasonably extended?

src/clashing_inputs.rs
Shute052: This is about handling input conflicts. If I understand correctly, when Input A and Input B conflict, one of them is a true subset of the other.

`Group` is used for Chords and other similar input combinations (like `Sequences` in future), and these are made up of one or more inputs.
`Composite` is for inputs that are composed of at least two inputs, like `VirtualAxis` and `VirtualDPad`, which are made up of two and four buttons, respectively.

But maybe it could be changed to return a Vec and just check the lengths and basic inputs of each.

src/raw_inputs.rs
Shute052: This struct has just been moved out from the original user_input.rs file.

src/user_input/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+///
+/// // Remember to register your input - it will ensure everything works smoothly!
+/// let mut app = App::new();
+/// app.register_user_input::<MouseScrollAlwaysFiveOnYAxis>();
+/// ```
+pub trait UserInput:
+    Send + Sync + Debug + DynClone + DynEq + DynHash + Reflect + erased_serde::Serialize
+{
+    /// Defines the kind of behavior that the input should be.
+    fn kind(&self) -> InputControlKind;
```
alice-i-cecile, line 209: I think that the better design here is to have three separate traits `ButtonlikeUserInput` etc that all rely on a core `UserInput` trait that captures shared behavior. I think that's likely to be easier to do in a follow-up though.

src/clashing_inputs.rs
Shute052: This is the old `UserInput::len`

alice-i-cecile (commented):
Okay so I've finally had a chance to review this in earnest. Overall, I'm very impressed with the quality of both the engineering and documentation: thank you very very much.

The net effect of this refactor is great: much more flexible and principled, with some of the code that I hated most (InputStream) gone.

I have a small concern, which should be addressed now, and then two larger ones, which may make sense to wait on:

1. I would like to avoid breaking the end user API when possible during this: can we keep `UserInput::chord` and `UserInput::modified`?
2. I don't like the `BasicInput` struct. I understand why it's needed for clash handling, but I think it's the wrong design because it means that custom implemented user input type won't get clashes resolved properly (and we're back to hardcoding the set).
3. The `UserInput` trait is too monolithic, and requires the implementation of methods that don't make sense for all input kinds. I think this should be split into buttonlike/single-axis/dual-axis.

src/user_input/mod.rs
Shute052: I had previously considered splitting it into different input kinds: Button, Axis, DualAxis, etc.; and there are corresponding data types for each. Then, modify the ActionState API, something like this:
```
if let ButtonState(state) = action_state.get(ButtonAction) {
   let pressed = state.pressed;
   let pressure = state.pressure; // Gamepad buttons are ranging from 0 to 1, and others are 0 or 1
}
if let AxisValue(value: f32) = action_state.get(AxisAction) {
}
```

Shute052:
> 1. I would like to avoid breaking the end user API when possible during this: can we keep `UserInput::chord` and `UserInput::modified`?



- Chord:
  - pub fn new<U: UserInput>(inputs: impl IntoIterator<Item = U>) -> Self
  - pub fn from_single(input: impl UserInput) -> Self
  - pub fn with(mut self, input: impl UserInput) -> Self
  - pub fn with_multiple<U: UserInput>(mut self, inputs: impl IntoIterator<Item = U>) -> Self
- Modifier:
  - pub fn with(&self, other: impl UserInput) -> InputChord

src/input_map.rs
Shute052: I think it might be a bit inconvenient because if you use impl IntoIterator<Item = impl UserInput> as a parameter, you know that the iterated elements must be of the same type due to the limitations of generics

src/raw_inputs.rs
Shute052: It should be able to replace with a new trait for implementing input mocking

alice-i-cecile:
Alright, I can get behind that. Let's get this merged and then we can start experimenting on further refinement.

alice-i-cecile (approved):

src/raw_inputs.rs
alice-i-cecile: Oh, I see!

src/clashing_inputs.rs
alice-i-cecile: Ah, got it.

alice-i-cecile:
Adopted!

PR #494, opened 2024-02-28
Replace axis-like input configuration with new input processors

### LINKED ISSUE #491: New Input Processors
Add the input processors, similar to those in [Unity](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/manual/Processors.html), and replace the existing ones

## Processor Types

- Clamp
- Filter: for all values, only negative values, only positive values, or no passing values
- Sensitivity
- Inversion
- Dead Zone: circle, square, rounded square

### LINKED ISSUE #495: Sensitivity applied before deadzone for dual axis inputs
## Version

0.13.3

## Operating system & version

NixOS unstable

## What you did

Set input sensitivity <= deadzone, i.e.: 

```
DualAxis::right_stick()
            .with_sensitivity(0.1, 0.1)
            .with_deadzone(DeadZoneShape::Ellipse {
                radius_x: 0.1,
                radius_y: 0.1,
            }),
```

Non-zero input values.

## What actually happened

Zeroes.

## Additional information

If the sensitivity for a dual axis input is less than or equal to the thresholds, the axis will never be considered "pressed". It seems `extract_dual_axis_data` ignores the deadzone for each axis (as it should) but applies the sensitivity, then checks the deadzone later. In the aforementioned case the inputs are never outside the deadzone. I think this is a bug. The sensitivity should be applied after the deadzone is checked?

Shute052:
# Objective

- Parts of #491 to replace the existing ones with new input processors.
- Fixes #495.

## Solution

Added input processors for `SingleAxis`, `DualAxis`,  `VirtualAxis`, and `VirtualDpad` to refine input values:
  - Traits:
    - `AxisProcessor`: Handles single-axis values.
    - `DualAxisProcessor`: Handles dual-axis values.
  - Built-in Processors:
    - Pipelines: Combine multiple processors into a pipeline.
      - `AxisProcessingPipeline`: Chain processors for single-axis values.
      - `DualAxisProcessingPipeline`: Chain processors for dual-axis values.
    - Inversion: Reverses control (positive becomes negative, etc.)
      - `AxisInverted`: Single-axis inversion.
      - `DualAxisInverted`: Dual-axis inversion.
    - Sensitivity: Adjusts control responsiveness (doubling, halving, etc.).
      - `AxisSensitivity`: Single-axis scaling.
      - `DualAxisSensitivity`: Dual-axis scaling.
    - Value Bounds: Define the boundaries for constraining input values.
      - `AxisBounds`: Restricts single-axis values to a range.
      - `DualAxisBounds`: Restricts single-axis values to a range along each axis.
      - `CircleBounds`: Limits dual-axis values to a maximum magnitude.
    - Deadzones: Ignores near-zero values, treating them as zero.
      - Unscaled versions:
        - `AxisExclusion`: Excludes small single-axis values.
        - `DualAxisExclusion`: Excludes small dual-axis values along each axis.
        - `CircleExclusion`: Excludes dual-axis values below a specified magnitude threshold.
      - Scaled versions:
        - `AxisDeadZone`: Normalizes single-axis values based on `AxisExclusion` and `AxisBounds::default`.
        - `DualAxisDeadZone`: Normalizes dual-axis values based on `DualAxisExclusion` and `DualAxisBounds::default`.
        - `CircleDeadZone`: Normalizes dual-axis values based on `CircleExclusion` and `CircleBounds::default`.
   - App extensions: `register_axis_processor` and `register_dual_axis_processor` for registering processors in the `bevy::prelude::App`.
   - Procedual macro attribute: `serde_typetag` for adding type tags for processors.

## TODO

- [x] Add new examples for processors.
- [x] Add new examples for inputs using processors.

## Known Issues

- Deserializing generic trait objects directly in Rust presents a current limitation. As a workaround, we need to define a corresponding concrete type for each desired value type used with the trait.

## Changelog

- Replace `DualAxisShape` with new deadzones.
- Replace functions for inversion, sensitivity, and deadzone creation with processor constructors.

### Internal changes

- Add dependency to `typetag`, `dyn_clone`, `dyn_eq`, `dyn_hash` for boxed dyn trait objects, ensuring object-safety.
- Refactor the logic of `InputStreams::input_value` to use these processors.


Shute052:
the dead zone seems to be a bit... dead

src/input_processing/dual_axis.rs
```diff
@@ -0,0 +1,1544 @@
...
+        // Short-circuiting if the input value should be excluded.
+        if self.exclusion.contains(input_value) {
+            return Vec2::ZERO;
+        }
+
+        // Short-circuiting if the input value should be clamped.
+        if !self.bounds.contains(input_value) {
+            return self.bounds.process(input_value);
+        }
+
```
Shute052, line 945: I reviewed the generated assembly code and observed that the compiler smartly inlined these functions and removed redundant code segments, resulting in an outcome identical to manually inlining the code.

Cargo.toml
```diff
@@ -50,6 +50,10 @@ derive_more = { version = "0.99", default-features = false, features = [
 ] }
 itertools = "0.12"
 serde = { version = "1.0", features = ["derive"] }
+typetag = "0.2"
```
alice-i-cecile, line 53: This won't work on WASM: we need to find an alternate solution.

examples/axis_inputs.rs
```diff
@@ -33,9 +33,10 @@ fn spawn_player(mut commands: Commands) {
         .insert(Action::Throttle, GamepadButtonType::RightTrigger2)
         // And we'll use the right stick's x-axis as a rudder control
         .insert(
+            // Add an AxisDeadzone to process horizontal values of the right stick.
             // This will trigger if the axis is moved 10% or more in either direction.
             Action::Rudder,
-            SingleAxis::symmetric(GamepadAxisType::RightStickX, 0.1),
+            SingleAxis::new(GamepadAxisType::RightStickX).with_processor(AxisDeadzone::default()),
```
alice-i-cecile, line 39: This will be clearer and more robust if we specify the actual value, rather than relying on the `Default` value aligning.

examples/input_processing.rs
```diff
@@ -0,0 +1,70 @@
...
+            VirtualDPad::wasd()
+                // You can add a processor to handle axis-like user inputs by using the `with_processor`.
+                //
+                // This processor is a circular deadzone that normalizes input values
+                // by clamping their magnitude to a maximum of 1.0,
+                // excluding those with a magnitude less than 0.1,
+                // and scaling other values linearly in between.
+                .with_processor(CircleDeadzone::default())
+                // Followed by appending another processor for the next processing step.
+                .with_processor(AxisSensitivity(2.0).extend_dual()),
```
alice-i-cecile, line 36: `extend_dual` feels like boilerplate to me. Can we remove it with `impl Into` style tricks?

examples/mouse_position.rs
```diff
@@ -60,8 +60,7 @@ fn update_cursor_state_from_window(
 
             if let Some(val) = window.cursor_position() {
                 action_state
-                    .action_data_mut(&driver.action)
-                    .unwrap()
+                    .action_data_mut_or_default(&driver.action)
```
alice-i-cecile, line 63: We should make `action_data` and `action_data_mut` infallible again, and remove this. Seperate PR though; I don't care whether it's before or after this PR is merged.

examples/send_actions_over_network.rs
```diff
@@ -23,10 +23,6 @@ enum FpsAction {
     Shoot,
 }
 
-/// This identifier uniquely identifies entities across the network
-#[derive(Component, Clone, PartialEq, Eq, Hash, Debug)]
-struct StableId(u64);
```
alice-i-cecile, line 28: This cleanup is good, but not particularly related. Whatever, it can stay.

src/input_processing/dual_axis/bounds.rs
```diff
@@ -0,0 +1,299 @@
...
+
+use bevy::prelude::*;
+use bevy::utils::FloatOrd;
+use serde::{Deserialize, Serialize};
+
+use super::super::single_axis::*;
+use super::DualAxisProcessor;
+use crate::define_dual_axis_processor;
+
+define_dual_axis_processor!(
```
alice-i-cecile, line 13: No macros please. I don't think the need is compelling enough to justify the serious cost to maintainability.

src/input_processing/dual_axis/bounds.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    /// Checks whether the `input_value` is within this deadzone.
+    #[must_use]
+    #[inline]
+    pub fn contains(&self, input_value: Vec2) -> bool {
+        input_value.length_squared() <= self.radius_squared
+    }
+}
+
+impl Eq for CircleBounds {}
```
alice-i-cecile, line 159: This can be derived.

src/input_processing/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+//!
+//! Choose between pipelines and macros based on your specific processing needs and performance requirements.
+
+pub use self::dual_axis::*;
+pub use self::single_axis::*;
+
+pub mod dual_axis;
+pub mod single_axis;
+
+// region pipeline
```
alice-i-cecile, line 85:
```suggestion
```

src/input_processing/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+
+/// Defines a new pipeline called the `name` that processes `value_type` values sequentially
```
alice-i-cecile, line 87:
```suggestion
/// Defines a new pipeline called `name` that processes `value_type` values sequentially
```

examples/input_processing.rs
Shute052: Since the implementations generated by define_dual_axis_processor are enum variants (All, Separate(x, y), OnlyX, OnlyY), I feel like repeating code such as `DualAxisSensitivity::All(AxisSensitivity(2.0))` would be more cumbersome

src/input_processing/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+/// through a chain of specified `processors` implementing the `processor_type`.
+/// This enhances performance by enabling optimizations like inlining and dead code elimination.
```
alice-i-cecile, line 89: IMO this isn't worth the complexity. Performance requirements here are pretty minimal, and we don't have the benchmarks to justify this.

src/input_processing/single_axis.rs
```diff
@@ -0,0 +1,1084 @@
...
+use std::hash::{Hash, Hasher};
+
+use bevy::prelude::*;
+use bevy::utils::FloatOrd;
+use dyn_clone::DynClone;
+use dyn_eq::DynEq;
+use dyn_hash::DynHash;
+use serde::{Deserialize, Serialize};
+
+// region processor trait
```
alice-i-cecile, line 13:
```suggestion
```

src/input_processing/single_axis.rs
```diff
@@ -0,0 +1,1084 @@
...
+    /// Processes the `input_value` and returns the result.
+    fn process(&self, input_value: f32) -> f32;
+}
+
+dyn_clone::clone_trait_object!(AxisProcessor);
+dyn_eq::eq_trait_object!(AxisProcessor);
+dyn_hash::hash_trait_object!(AxisProcessor);
+crate::__reflect_box_dyn_trait_object!(AxisProcessor);
+
+// endregion processor trait
```
alice-i-cecile, line 89:
```suggestion
```

src/input_processing/single_axis.rs
```diff
@@ -0,0 +1,1084 @@
...
+
+// region pipeline
```
alice-i-cecile, line 91:
```suggestion
```

src/input_processing/single_axis.rs
```diff
@@ -0,0 +1,1084 @@
...
+    fn merge_processor(self, processor: impl AxisProcessor) -> Self {
+        let pipeline = match Reflect::as_any(&*self).downcast_ref::<AxisProcessingPipeline>() {
+            Some(pipeline) => pipeline.clone(),
+            None => AxisProcessingPipeline(vec![self]),
+        };
+        Box::new(pipeline.with(processor))
+    }
+}
+
+// endregion pipeline
```
alice-i-cecile, line 151:
```suggestion
```

src/input_processing/single_axis.rs
```diff
@@ -0,0 +1,1084 @@
...
+
+// region inversion
```
alice-i-cecile, line 153:
```suggestion
```

src/input_processing/single_axis.rs
```diff
@@ -0,0 +1,1084 @@
...
+    #[inline]
+    fn process(&self, input_value: f32) -> f32 {
+        -input_value
+    }
+}
+
+// endregion inversion
+
+// region sensitivity
+
```
alice-i-cecile, line 187:
```suggestion
```

src/input_processing/mod.rs
Shute052: image

In IDEs and vscode, `// region` can be used to collapse code sections, but now that the file length is not long, it can be removed

src/axislike.rs
```diff
@@ -20,139 +20,123 @@ use serde::{Deserialize, Serialize};
...
 
-    /// Returns this [`SingleAxis`] inverted.
-    #[must_use]
-    pub fn inverted(mut self) -> Self {
-        self.inverted = !self.inverted;
-        self
+    /// Get the "value" of this axis.
+    /// If a processor is set, it will compute and return the processed value.
+    /// Otherwise, pass the `input_value` through unchanged.
+    #[inline]
```
alice-i-cecile, line 128:
```suggestion
    #[inline]
    #[must_use]
```

Cargo.toml
Shute052: I checked and it seems there is a crate called `bevy_serde_project` that reimplements `typetag` for WASM, but I'm not sure if it's really usable, as it's still in version 0.1

alice-i-cecile (changes requested):
Generally quite pleased with this abstraction, and the builder pattern used! It makes a lot of sense to me. Code quality and docs are exceptional, you've done an awesome job there.

However:

1. The pipeline work is super cool, but not needed for an MVP. I'm opposed to merging it in our initial implementation, and will require benchmarks to merge it in a followup. Simply storing a `Vec<Box<dyn AxisProcessor>>` is a fine design to start with.
2. Some cleanup to do: the `region` comments should all be gone.
3. Fewer macros. The `Reflect` for trait objects one can stay, but the others should be cut. They're relatively hard to maintain, and will be a serious barrier to upstreaming into Bevy.
4. `typetag` cannot be used, flat out. It's incompatible with WASM builds, so we need to redesign to strip it out. See https://github.com/rustwasm/wasm-bindgen/issues/1216.

alice-i-cecile:
The processors still need to be removed from this PR. I also don't quite understand why all of the typetag stuff is needed in the first place. Is this just for serialization of processors? I would really like to avoid introducing the associated complexity and dependencies, and would frankly rather sacrifice the ability to store arbitrary preprocessor functions than add it.

I understand if that's not the path you'd prefer to take: I'm happy to adopt this excellent work and finish it up on my own if you'd like.

alice-i-cecile (approved):
Okay, I'm satisfied now. To rehash from our conversation [on Discord](https://discord.com/channels/691052431525675048/1034547742262951966/1228411354872807525):

1. Pipelines are much simpler now, and just store a `Vec<Boxy dyn Processor>`.
2. We need typetag-style functionality for serialization of boxed trait objects. While we *could* move away from trait objects here, in favor of a finite set of operations, it won't actually trim complexity / dependencies in the end.

The case for using trait objects for input types is dramatically stronger, as is the need for serialization.

Merging!

PR #499, opened 2024-03-15
docs: add multiplayer action usage example

chaosteil:
## What was the problem?

When reading the multiplayer example it wasn't entirely clear how to query for
each player state directly, as a lot of the examples the the ActionState as a
resource, thus making it a singleton. Adding this usage pattern in the example
makes it just a bit clearer.

## How did you fix it?

I added a system that uses the actions that were provided in the setup system in
the multiplayer example.


alice-i-cecile (approved):

alice-i-cecile:
Thanks! I agree that this is a useful extension to the example.

PR #501, opened 2024-03-22
Fixes a bug where gamepads are not filtered

### LINKED ISSUE #500: Gamepad actions are sent to all related `ActionState` components, regardless of restrictions placed by `InputMap::set_gamepad()`
## Version

Release 0.13.3

## Operating system & version

* Kubuntu 23.10 x86_64, with Kernel 6.5.0-26-generic

## What you did

Requires the use of 2 gamepads. 

0. Clone the repository and 
1. Execute `cargo run --example register_gamepads`. 
2. Once the window opens, register both controllers with L+R bumpers.
3. Press the south face button on either controller.

The log should display a single instance of the "Player # jumped!" message (where '#' is number of the gamepad used).

## What actually happened

The log instead displays the message a number of times equal to the number of registered gamepads. 

## Additional information

This seems to result from a deeper problem. The function `InputMap::set_gamepad(..)` is intended to restrict the gamepads associated with an InputMap and ActionState component bundle, but it seems to have no impact in this case.

EmmiaSilk:
Fixes #500. I believe the issue was caused by the `InputStreams::button_pressed()` function. I rewrote the associated block to filter out unassociated gamepads, but also to let through any gamepad if no associated gamepad is defined. 

I make no guarantees regarding if this is a well-written solution. Feedback is appreciated!

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -6,6 +6,10 @@
 
 - Removed `Direction` type. Use `bevy::math::primitives::Direction2d`.
 
+### Bugs
+
+- fixed a bug in `InputStreams::button_pressed()` where unrelated gamepads were not filtered out when an `associated_gamepad` is defined.
+
 ## Version 0.13.3
 
 ### Bugs
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -6,6 +6,10 @@
 
 - Removed `Direction` type. Use `bevy::math::primitives::Direction2d`.
 
+### Usability
+
+- added `InputMap::set_exclusive_gamepad()` that restricts input to the specified [`Gamepad`] for the entity controlled by this input map.
+
 ### Bugs
 
 - fixed a bug in `InputStreams::button_pressed()` where unrelated gamepads were not filtered out when an `associated_gamepad` is defined.
--- a/examples/register_gamepads.rs
+++ b/examples/register_gamepads.rs
@@ -49,7 +49,9 @@ fn join(
                     (Action::Disconnect, GamepadButtonType::Select),
                 ])
                 // Make sure to set the gamepad or all gamepads will be used!
-                .set_gamepad(gamepad)
+                // If you don't want a specified exclusive gamepad,
+                // use `set_gamepad()` instead.
+                .set_exclusive_gamepad(gamepad)
                 .build();
                 let player = commands
                     .spawn(InputManagerBundle::with_map(input_map))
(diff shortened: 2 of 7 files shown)


Shute052:
Perhaps we can change the `associated_gamepad` in `InputMap` from `Option<Gamepad>` to an enum that represents multiple handling options like:

```rust
pub enum GamepadValueSource {
    Any,
    Only(Gamepad),
    Prefer(Gamepad),
}
```

Shute052:
Do you need any guidance on gamepad processing, or should I create a PR to your branch?

EmmiaSilk:
I think to start with, it may be best to make a pull request onto my branch.

alice-i-cecile:
Fully on board with that direction :)

Shute052:
Created the [PR](https://github.com/EmmiaSilk/leafwing-input-manager/pull/1)

EmmiaSilk:
Finally had the opportunity to verify controller behaviors-- The `set_exclusive_gamepad()` function feels like a nice solution to not break the the API.

alice-i-cecile (commented):
I like the shape of this solution: can you add tests to prevent regressions?

EmmiaSilk:
I will begin writing a suite of tests for this functionality.

EmmiaSilk:
I came upon a problem in building my tests, as I am not certain of the intended behavior of `GamepadValueSouce::Prefer`. It seems obvious to me that, in cases where an input is received from more than one gamepad, the plugin should prefer one gamepad over the other. However, there may arise cases in which one gamepad provides input (such as a "Press" on Gamepad 1), then on the next frame another gamepad provides an opposing input (such as a "release" on Gamepad 2). 

In these circumstances, should I expect the system to respect the high state of the preferred gamepad, or should I expect it to respect the Gamepad whose state changed most recently?

Shute052:
Ahh, if I'm not mistaken, `Prefer` uses the original code logic. If the specified gamepad doesn't have input, it fetches from others that are registered.

Maybe it should be changed to fetch from elsewhere only if it's not functioning properly, or add a new variant?

alice-i-cecile:
I think that `Prefer` is probably just a footgun: I don't see how that behavior would ever be useful.

Shute052:
Oh no, it seems like there was a misunderstanding on my part, which led to the bug introduced in #466. The old code was actually correct.

Shute052:
Closed since #504 merged

PR #504, opened 2024-03-30
Fixed gamepad filtering (alternate branch)

### LINKED ISSUE #500: Gamepad actions are sent to all related `ActionState` components, regardless of restrictions placed by `InputMap::set_gamepad()`

EmmiaSilk:
Fixes #500 (again), as an alternative to the solution provided in #501.
This also adds a new test group (`tests/multiple_gamepads.rs`) and (hopefully) clarifies the documentation for `InputMap.set_gamepad()`.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -6,6 +6,10 @@
 
 - Removed `Direction` type. Use `bevy::math::primitives::Direction2d`.
 
+### Bugs
+
+- fixed a bug in `InputStreams::button_pressed()` where unrelated gamepads were not filtered out when an `associated_gamepad` is defined.
+
 ## Version 0.13.3
 
 ### Bugs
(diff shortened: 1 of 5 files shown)


alice-i-cecile (approved):
Great stuff, thank you!

PR #514, opened 2024-05-01
Add `WithProcessorExt` traits and impls for short input processing configuration

Shute052:
- Should be merged once #513 was merged.
- Added two traits: `WithAxisProcessorExt` and `WithDualAxisProcessorExt`.
- Implemented them for `SingleAxis`, `DualAxis`, `VirtualAxis`, and `VirtualDPad`.

PR #515, opened 2024-05-03
Remove the hacky `value` field and `from_value` method of axis-like inputs in favor of a new input mocking API

Shute052:
- should be merged once #514 was merged.
- added new methods for sending value changed events to each axis within the input:
  - `fn send_axis_values(&self, input: UserInput, values: impl IntoIterator<Item = f32>)`.
  - `fn send_axis_values_as_gamepad(&self, input: UserInput, values: impl IntoIterator<Item = f32>, gamepad: Option<Gamepad>)`.
- renamed `MockInput::send_input` method to `MockInput::press_input` to better reflect its current functionality for clarity.
- removed the hacky `value` field and `from_value` method from `SingleAxis` and `DualAxis`, in favor of new input mocking API.

# Migration Guide

If you were using the `send_input` method for buttons and keys to send simulated pressed events, simply update the method name to `press_input`.

If you were using the `value` field or the `from_value` method to create a `SingleAxis` or `DualAxis` instance for sending simulated input events, you can remove them entirely.
Simply use a normal version of these inputs and call the `send_axis_values` method with your desired values.

### DIFF
--- a/src/input_mocking.rs
+++ b/src/input_mocking.rs
@@ -1,116 +1,192 @@
 //! Helpful utilities for testing input management by sending mock input events
 //!
 //! The [`MockInput`] trait contains methods with the same API that operate at three levels:
-//! [`App`], [`World`] and [`MutableInputStreams`], each passing down the supplied arguments to the next.
 //!
-//! Inputs are provided in the convenient, high-level [`UserInput`] form.
-//! These are then parsed down to their [`UserInput::raw_inputs()`],
-//! which are then sent as [`bevy::input`] events of the appropriate types.
+//! 1. [`App`].
+//! 2. [`World`].
+//! 3. [`MutableInputStreams`].
+//!
+//! Each passing down the supplied arguments to the next.
 
-use crate::axislike::{AxisType, MouseMotionAxisType, MouseWheelAxisType};
-use crate::buttonlike::{MouseMotionDirection, MouseWheelDirection};
-use crate::input_streams::{InputStreams, MutableInputStreams};
-use crate::user_input::{RawInputs, UserInput};
+use bevy::ecs::system::SystemState;
+use bevy::input::gamepad::{Gamepad, GamepadButton, GamepadButtonType, GamepadEvent};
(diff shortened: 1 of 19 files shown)


alice-i-cecile:
Oh thank goodness I hated that design!

Shute052:
> Oh thank goodness I hated that design!谢天谢地，我讨厌这个设计。

Hmm, which way is better? I feel like adding a value field to the input is even stranger.

src/input_mocking.rs
```diff
@@ -1,116 +1,190 @@
...
     ///
-    /// Note that inputs will continue to be pressed until explicitly released or [`MockInput::reset_inputs`] is called.
+    /// # Limitations
     ///
-    /// Provide the [`Gamepad`] identifier to control which gamepad you are emulating.
-    fn send_input_as_gamepad(&mut self, input: impl Into<UserInput>, gamepad: Option<Gamepad>);
+    /// Non-binary inputs (e.g., analog thumb sticks) aren't affected.
+    /// Use [`Self::send_axis_values`] for those.
+    ///
+    /// Unfortunately, due to upstream constraints,
```
alice-i-cecile, line 99: Link to a Bevy issue here please. If none exist, make one then link it.

src/input_processing/single_axis/mod.rs
```diff
@@ -165,6 +165,81 @@ impl Hash for AxisProcessor {
     }
 }
 
+/// Provides methods for configuring and manipulating the processing pipeline for single-axis input.
+pub trait WithAxisProcessorExt: Sized {
```
alice-i-cecile, line 169: Does this need to be rebased? This should have been added already in #514.

alice-i-cecile:
Broadly in favor of the refactor: I like how this looks, and the split is cleaner. The diff is quite messy still: can you rebase?

alice-i-cecile (approved):
Good, I like these changes too. Splitting out the different forms of input mocking is much cleaner.

Issue #525, opened 2024-05-24
Put the 'Timing' field of 'ActionState' behind a feature


cBournhonesque:
## What problem does this solve?

There are a lot of games that don't need information about timings, but simply about button presses.
It would improve performance + bandwidth (for networking) if the timing fields were put behind a feature.

If the feature is off, the field wouldn't get any timing information, and it would not get updated at all.

alice-i-cecile:
Agreed! I would happily merge a PR for this.

PR #526, opened 2024-05-24
Add timing data behind a feature

cBournhonesque:
Fixes https://github.com/Leafwing-Studios/leafwing-input-manager/issues/525


alice-i-cecile (changes requested):
Needs a release note entry :) Make sure to note that this is off-by-default. And double check that the relevant example still works with the current feature config.

alice-i-cecile (approved):

PR #545, opened 2024-06-12
Remove duplicate Hash bound for Actionlike

msvbg:
A trivial/harmless problem: `Hash` occurs twice in the trait bounds.


alice-i-cecile (approved):
Huh, interesting that that's not a compiler error or lint 🤔

PR #548, opened 2024-06-22
Refactor `UserInput` into `Buttonlike`, `Axislike` and `DualAxislike`

### LINKED ISSUE #535: Split UserInput into multiple traits
I had previously considered splitting it into different input kinds: Button, Axis, DualAxis, etc.; and there are corresponding data types for each. Then, modify the ActionState API, something like this:
```
if let ButtonState(state) = action_state.get(ButtonAction) {
   let pressed = state.pressed;
   let pressure = state.pressure; // Gamepad buttons are ranging from 0 to 1, and others are 0 or 1
}
if let AxisValue(value: f32) = action_state.get(AxisAction) {
}
```

_Originally posted by @Shute052 in https://github.com/Leafwing-Studios/leafwing-input-manager/pull/490#discussion_r1632406354_

alice-i-cecile:
Fixes #535.

This is a major refactor that splits user inputs into three kinds: buttonlike, axislike and dualaxislike.
This change ends up rippling through the whole stack: from `Actionlike` to `UserInput` to `InputMap` to `ActionState`.

Despite that, I'm really happy with this change! It really clarifies the semantics of the different input types, in a way that I think will be easier to learn, more natural to maintain and harder to mess up.

There are a number of items that need to be done before this is mergeable still:

- [x] rename with_dualaxis to with_dual_axis
- [x] generate `ActionDiff`s correctly
- [x] sanity-check clashing inputs
- [x] actually update axislike and dualaxislike data in `ActionState::update`'s call stack
- [x] fix the lengths of all of our different user input kinds
- [x] go over `ActionState` to double-check that all APIs have been updated correctly
- [x] make everything compile
- [x] cargo fmt
- [x] cargo clippy
- [x] get tests passing 🥲 

## Follow-up

This PR is already way, way too big. In the interest of sanity, some functionality has been omitted / left broken in this PR.

- [ ] store 0-1 values in Buttonlike inputs to properly handle triggers
- [ ] fix clash handling for dpads
- [ ] fix action diffing


src/action_state.rs
```diff
@@ -125,70 +169,149 @@ impl ActionData {
 /// ```
 #[derive(Resource, Component, Clone, Debug, PartialEq, Serialize, Deserialize, Reflect)]
 pub struct ActionState<A: Actionlike> {
-    /// The [`ActionData`] of each action
-    action_data: HashMap<A, ActionData>,
+    /// The [`ButtonData`] of each action
+    button_data: HashMap<A, ButtonData>,
+    /// The [`AxisData`] of each action
+    axis_data: HashMap<A, AxisData>,
```
cBournhonesque, line 175: Is it intended that an action could be associated with both an AxisData and a ButtonData?

cBournhonesque: Oh, Button-presses, 1-d joysticks, and 2-d joysticks are now completely independent?

alice-i-cecile: Yes, they are. Generally you won't reasonably end up with an action associated with more than one of the input types though.

cBournhonesque: I wonder if some metadata should be shared between all 3 options.
For example the `ActionState` being disabled seems more-or-less independent from the way the action is mapped (button/axis/axis-2d).

For example in my case I want to stop input replication systems if the Action is disabled. Now it looks like i would have to check independently for all 3 subtypes of the action

alice-i-cecile: Right 🤔 I didn't think about that super hard, but I suspect that we should be able to refactor for that... I think probably in a follow-up: I want to get this PR out the door before it grows any larger.

PR #586, opened 2024-08-09
Reflect InputMap and ActionState

ChristopherBiscardi:
blenvy requires the ReflectComponent data when instantiating blueprints.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,17 @@
 # Release Notes
 
+## Version 0.15.1
+
+### Usability (0.15.1)
+
+#### InputMap
+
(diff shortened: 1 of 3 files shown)


alice-i-cecile (changes requested):
Can you add `reflect(Resource)` as well as part of this? These are also usable as resources and I don't want to forget.

alice-i-cecile (approved):

alice-i-cecile:
Actually, one more thing: can you update RELEASES.md, noting this down as a usability fix?

ChristopherBiscardi:
@alice-i-cecile I assumed this would be in a new 0.15.1 section, so I added the relevant headings. Can change if that's incorrect.

PR #601, opened 2024-08-19
Add ability to filter entities for which action diffs should be generated

### LINKED ISSUE #600: Generate action diffs for only a filtered set of entities

In a networked game, there may be cases where actions on some entities should remain local, while actions on others should get diffed and have `ActionDiffEvent<A>`s generated.

Currently, `SummarizedActionState<A>::summarize` & `generate_action_diffs<A>` summarize and emit action diffs for every entity with `ActionState<A>`, which lacks the degree of control I'm looking for.

I'd like to be able to filter which entities get action diffs generated.

Since we need to add the `generate_action_diffs<A>` system manually anyway, this seems to be the best place to specify it, e.g. `generate_action_diffs<A, Without<NotActionSender<A>>` (where `NotActionSender<A>` is an example of a tag component that marks anything for which action diff events should not be sent).

Implement a new `generate_action_diffs_filtered<A, F = QueryFilter>` function, and have `generate_action_diffs<A>` call it with `F = ()`. User code can choose between adding the `generate_action_diffs<A>` or `generate_action_diffs_filtered<A, F>` systems according to their needs.

See PR #601.

## What alternatives have you considered?

- Require user code to use different Actionlikes to differentiate actions that should be diffed and those that shouldn't, and only call `generate_action_diffs` on the former. It's a pretty big burden on the user side though.
- Create a new tag component in this crate (e.g. `NotActionSender<A>`, or a better name) that users can tag entities with, and have `generate_action_diffs<A>` always ignore entities with this tag. A bit less flexible.

zhaop:
Fixes #600.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -2,6 +2,13 @@
 
 ## Version 0.15.1
 
+### Enhancements (0.15.1)
+
+#### Action diffs on filtered entities
+
+- Added the `generate_action_diffs_filtered<A, F>` system, which accepts a `QueryFilter`, so that only entities matching QueryFilter `F` (and with `ActionState<A>`) generate action diffs.
+ - To support this, `SummarizedActionState<A>` also gets a new `summarize_filtered<F>` function alongside the original `summarize`.
+
 ### Usability (0.15.1)
 
 #### InputMap
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -9,8 +9,9 @@
   - removed `KeyboardVirtualDPad` and `GamepadVirtualDPad` in favor of `VirtualDPad`
   - removed `KeyboardVirtualDPad3D` in favor of `VirtualDPad3D`
 - added `threshold` value for `GamepadControlDirection`, `MouseMoveDirection`, and `MouseScrollDirection` to be considered pressed.
-- Added the `generate_action_diffs_filtered<A, F>` system, which accepts a `QueryFilter`, so that only entities matching QueryFilter `F` (and with `ActionState<A>`) generate action diffs.
-  - To support this, `SummarizedActionState<A>` also gets a new `summarize_filtered<F>` function alongside the original `summarize`.
+- added ability to filter entities to generate action diffs for:
+  - added new `generate_action_diffs_filtered<A, F>` system, which accepts a `QueryFilter`, so that only entities matching QueryFilter `F` (and with `ActionState<A>`) generate action diffs
+  - added new `summarize_filtered<F>` function for `SummarizedActionState<A>` (alongside the original `summarize`).
 
 ## Version 0.15.1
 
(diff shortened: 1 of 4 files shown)


alice-i-cecile:
A nice idea! Remember to update RELEASES.md

src/action_diff.rs
```diff
@@ -103,9 +104,9 @@ impl<A: Actionlike> SummarizedActionState<A> {
     }
 
     /// Captures the raw values for each action in the current frame.
-    pub fn summarize(
+    pub fn summarize<F: QueryFilter>(
```
alice-i-cecile, line 107: Are we allowed to put a default type here?

We should add a non-filtering `F` as the default if so, and document what one is if not.

zhaop: I'm new to Rust, would a default type param look like `pub fn summarize<F: QueryFilter = ()>`? That gives me:

```
defaults for type parameters are only allowed in `struct`, `enum`, `type`, or `trait` definitions
this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
for more information, see issue #36887 <https://github.com/rust-lang/rust/issues/36887>
`#[deny(invalid_type_param_default)]` on by default
```

If it's not allowed, then let's rename the existing filtered implementation `summarize_filtered<F: QueryFilter>(...)`, and add a new `summarize(...)` that calls `summarize_filtered<()>()`.

alice-i-cecile: Yep, looks like it's not allowed. I agree with your suggestion :)

zhaop:
Okey I should be done! Let's see what CI thinks. I think with the changes to summarize this should no longer be a breaking change (so I put it in 0.15.1)?

RELEASES.md
```diff
@@ -2,6 +2,13 @@
 
 ## Version 0.15.1
 
+### Enhancements (0.15.1)
```
zhaop, line 5: Maybe more of a usability change?

alice-i-cecile: Agreed, I think it fits better there.

alice-i-cecile:
This also can't go in 0.15.1, since it's breaking. We can either refactor it to be completely non-breaking, or wait until 0.16

zhaop:
I just re-read about semantic versioning and agree 0.16 is a better place for this. Added a new 0.16 section to RELEASES.md.

To ensure this PR doesn't go into 0.15.1, should it remain unmerged until after 0.15.1 is released? or do we make a branch for 0.16?

alice-i-cecile:
Because this is nicely isolated I'll just leave this unmerged for a bit. I mostly intend to hit examples and trigger-like inputs for 0.15.1, neither of which should conflict.

alice-i-cecile:
@zhaop I've released 0.15.1 and would like to get this merged :) Can you resolve merge conflicts for me?

zhaop:
@alice-i-cecile Sure, done!

PR #603, opened 2024-08-22
Add `#[actionlike]` for actions to set their `InputControlKind`s

### LINKED ISSUE #587: Update the `Actionlike` derive macro to easily specify the types

Currently if the actions are not all Buttons, we need to implement ActionLike manually:
```
#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Copy, Hash, Reflect)]
pub enum PlayerActions {
    Up,
    Down,
    Left,
    Right,
    Shoot,
    MoveCursor,
}

impl Actionlike for PlayerActions {
    // Record what kind of inputs make sense for each action.
    fn input_control_kind(&self) -> InputControlKind {
        match self {
            Self::MoveCursor => InputControlKind::DualAxis,
            _ => InputControlKind::Button,
        }
    }
}
```

Have a derive macro to help out:

```
#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Copy, Hash, Reflect)]
pub enum PlayerActions {
    Up,
    Down,
    Left,
    Right,
    Shoot,
    #[actionlike(DualAxis)]
    MoveCursor,
}
```

Shute052:
- Fixes #587

# Solution

Added an `#[actionlike]` attribute to enhance ergonomics.

You can apply this attribute to the entire enum to set a default behavior for all action variants, or to individual variants to override the default behavior.

By default, all actions are treated as button-like, as they were previously.

```rust
#[derive(Actionlike, Debug, PartialEq, Eq, Clone, Copy, Hash, Reflect)]
#[actionlike(Axis)] // All variants in this enum are controlled by axes
enum CameraAction {
    Zoom,  // This action is controlled by axes
    #[actionlike(DualAxis)]
    Move,  // This action is controlled by dual axes since we have overrode the default option
    #[actionlike(Button)]
    TakePhoto, // This action is controlled by buttons since we have override the default option
}
```

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -8,6 +8,10 @@
 
 - Reflect `Component` and `Resource`, which enables accessing the data in the type registry
 
+#### Actionlike
+
+ - added `#[actionlike]` for actions to set their input kinds, either on an enum or on its individual variants.
+
 #### ActionState
 
 - Reflect `Component` and `Resource`, which enables accessing the data in the type registry
--- a/examples/axis_inputs.rs
+++ b/examples/axis_inputs.rs
@@ -14,21 +14,13 @@ fn main() {
         .run();
 }
 
-#[derive(PartialEq, Eq, Clone, Copy, Hash, Debug, Reflect)]
+#[derive(Actionlike, PartialEq, Eq, Clone, Copy, Hash, Debug, Reflect)]
 enum Action {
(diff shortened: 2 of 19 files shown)


alice-i-cecile (approved):
Perfect! This is exactly the behavior I was hoping for :D

PR #609, opened 2024-08-25
Add `TripleAxislike`

Shute052:
- added `TripleAxislike` trait for inputs that track all X, Y, and Z axes.
  - added `KeyboardVirtualDPad3D` that consists of six `KeyCode`s to represent a triple-axis-like input.
  - added `TripleAxislikeChord` that groups a `Buttonlike` and a `TripleAxislike` together.
  - added related variants such as:
    - `InputControlType::TripleAxis`
    - `ActionDiff::TripleAxisChanged`

PR #615, opened 2024-08-25
Add Triggerlike inputs

### LINKED ISSUE #551: Add Triggerlike inputs that have a 0 to 1 value

I want to properly model the behavior of buttons that have a value from 0 to 1, like the triggers on Joysticks.

Add a `Triggerlike` type (which implies `Buttonlike` and thus `UserInput`) which store this data.

## Related work

Cut from #548 for scope.

alice-i-cecile:
Fixes #551. Very similar to #609.

The basic infrastructure should be in place but this still needs:

- [ ] an integration test
- [ ] an example

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -28,7 +28,7 @@
 
 - added `#[actionlike]` for actions to set their input kinds, either on an enum or on its individual variants.
 
-#### ActionState
+#### ActionState reflection
 
 - Reflect `Component` and `Resource`, which enables accessing the data in the type registry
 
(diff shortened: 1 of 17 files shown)


RELEASES.md
```diff
@@ -9,15 +9,22 @@
...
   - added related variants such as:
     - `InputControlType::TripleAxis`
-    - `ActionDiff::TripleAxisChanged` 
+    - `ActionDiff::TripleAxisChanged`
+- added `Triggerlike` trait for buttonlike inputs whose values range from 0 to 1
+  - implemented `Triggerlike` for `GamepadButtonType`, `GamepadButton`, `KeyCode` and `MouseButton`
+  - typically, `GamepadButtonType` will be the `Triggerlike` used
+  - `KeyCode`, `MouseButton` and any non-trigger gamepad buttons will simply return 0 when released and 1 when pressed
+  - added related variants such as:
+    - `InputControlType::Triggerlike`
```
Shute052, line 18: This should be
```suggestion
    - `InputControlType::Trigger`
```

PR #621, opened 2024-09-10
Fix `InputMap::merge`

Shute052:
`InputMap::merge` is now compatible with all input kinds, previously limited to buttons


alice-i-cecile:
Remember to update the release notes :)

PR #622, opened 2024-09-10
Fix broken deserialization for inputs and `InputMap`s

Shute052:
- Fixed the broken deserialization that caused by #548
- Moved `#[serde_typetag]` attribute of inputs to correct places
- Changed `RegisterUserInput` extension for App
- Removed unnessary ser/de and reflect impls for `UserInput`
- Added a custom MapRegistry that allows multiple deserializer registration
- Added regression tests for ser/de of inputs and `InputMap`s


src/user_input/mod.rs
```diff
@@ -111,48 +111,7 @@ mod trait_serde;
...
-///         // This input is not buttonlike, so it uses `None`.
-///         BasicInputs::None
-///     }
-/// }
-///
-/// // Remember to register your input - it will ensure everything works smoothly!
-/// let mut app = App::new();
-/// app.register_user_input::<MouseScrollAlwaysFiveOnYAxis>();
-/// ```
-pub trait UserInput:
```
Shute052, line 153: I think we can add a detailed example.rs for this

alice-i-cecile: Agreed!

alice-i-cecile (approved):

PR #624, opened 2024-09-13
Add `threshold` value for `GamepadControlDirection`, `MouseMoveDirection`, and `MouseScrollDirection`

Shute052:
- Added a `threshold` field to `GamepadControlDirection`
- Renamed the `side` field of `GamepadControlDirection` to `direction` for clarity
- Changed the struct definitions of `MouseMoveDirection` and `MouseScrollDirection` from tuple structs to regular structs with `threshold` and `direction` fields

### DIFF
--- a/src/axislike.rs
+++ b/src/axislike.rs
@@ -26,7 +26,15 @@ impl AxisDirection {
     }
 
     /// Checks if the given `value` represents an active input in this direction,
-    /// considering the specified `threshold` (Must be non-negative).
+    /// considering the specified `threshold`.
+    ///
+    /// # Requirements
+    ///
+    /// - `threshold` >= `0.0`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the requirement isn't met.
     #[must_use]
     #[inline]
     pub fn is_active(&self, value: f32, threshold: f32) -> bool {
@@ -154,7 +162,15 @@ impl DualAxisDirection {
     }
 
     /// Checks if the given `value` represents an active input in this direction,
-    /// considering the specified `threshold` (Must be non-negative).
+    /// considering the specified `threshold`.
+    ///
+    /// # Requirements
+    ///
+    /// - `threshold` >= `0.0`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the requirement isn't met.
     #[must_use]
     #[inline]
     pub fn is_active(&self, value: Vec2, threshold: f32) -> bool {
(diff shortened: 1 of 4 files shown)


alice-i-cecile (changes requested):
This is breaking, and needs to be moved to 0.16 :) We're adding new fields to a public struct, and renaming one.

src/user_input/gamepad.rs
```diff
@@ -69,29 +72,45 @@ fn read_axis_value(
...
 pub struct GamepadControlDirection {
     /// The axis that this input tracks.
     pub axis: GamepadAxisType,
 
     /// The direction of the axis to monitor (positive or negative).
-    pub side: AxisDirection,
+    pub direction: AxisDirection,
+
+    /// The threshold value for the direction to be considered pressed.
+    pub threshold: f32,
```
uggla, line 85: :+1:

src/axislike.rs
```diff
@@ -25,13 +25,14 @@ impl AxisDirection {
...
+    /// Checks if the given `value` represents an active input in this direction,
+    /// considering the specified `threshold`.
     #[must_use]
     #[inline]
-    pub fn is_active(&self, value: f32) -> bool {
+    pub fn is_active(&self, value: f32, threshold: f32) -> bool {
         match self {
-            Self::Negative => value < 0.0,
-            Self::Positive => value > 0.0,
+            Self::Negative => value < -threshold,
```
uggla, line 34: I think I would use `-threshold.abs()`

src/axislike.rs
```diff
@@ -25,13 +25,14 @@ impl AxisDirection {
...
+            Self::Positive => value > threshold,
```
uggla, line 35: I think I would use:
```
Self::Negative => value < -threshold.abs(),
Self::Positive => value > threshold.abs(),
```

This could prevent user pitfalls.

src/axislike.rs
```diff
@@ -25,13 +25,14 @@ impl AxisDirection {
         }
     }
 
-    /// Checks if the given `value` represents an active input in this direction.
+    /// Checks if the given `value` represents an active input in this direction,
+    /// considering the specified `threshold`.
```
uggla, line 29: Maybe, it is worth mentioning that the threshold is a positive value.

uggla (commented):
I have mentioned a couple of minor changes that could prevent user pitfalls.
But overall, it looks good to me.

Shute052:
> We're adding new fields to a public struct, and renaming one.

Yeah, so in the previous PR, I leaned towards encapsulating those inputs

alice-i-cecile (approved):

Issue #648, opened 2024-10-14
Reading a button as an axis value no longer works


tec27:
## Version

0.15.1, or current main ( 5bb37ff7d07e472c0095a4b8fd60cbca6a1cd155 )

## Operating system & version

Windows 11

## What you did

1) Run the `axis_inputs` example (`cargo run --example axis_inputs`)
2) Press right trigger 2

## What you expected to happen

The program should print the value of the trigger press depth

## What actually happened

The program crashes:

```
thread 'Compute Task Pool (18)' panicked at C:\Users\Travis\.cargo\registry\src\index.crates.io-6f17d22bba15001f\leafwing-input-manager-0.15.1\src\action_state\mod.rs:594:14:
assertion `left == right` failed
  left: Button
 right: Axis
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `car_game::car::control_player`!
Encountered a panic in system `bevy_app::main_schedule::FixedMain::run_fixed_main`!
Encountered a panic in system `bevy_time::fixed::run_fixed_main_schedule`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
error: process didn't exit successfully: `target\debug\car-game.exe` (exit code: 101)
```

## Additional information

It looks like this assertion was added in: https://github.com/Leafwing-Studios/leafwing-input-manager/commit/5e90a7de5e438ea66704bf0eb90458528077591f

I'm not sure if the example needs to be updated to some alternative method (from looking through the docs I didn't see a great way of getting the button value at a higher level), or if this assertion isn't appropriate at this location.

alice-i-cecile:
This is a bug in the example, and intended to not work. The example should be checking the axis value explicitly.

crvarner:
@alice-i-cecile if you can point me in the right direction for how to check the axis explicitly I can try and fix the example

PR #649, opened 2024-10-16
Add `value` for button-like inputs

### LINKED ISSUE #551: Add Triggerlike inputs that have a 0 to 1 value

Shute052:
# Objective

- Fixes #551, #648
- Adopted #615

## Solution

- added `value` function and friends to `ButtonlikeInput` trait
- added `button_value` function and friends to `ActionState`
- added `value` field and friends to `action_data::ButtonData`
- `GamepadButton` and `GamepadButtonType` now get values from `Axis<GamepadButton>`
- `VirtualAxis`, `VirtualDPad`, and `VirtualDPad3D` now report axis values based on the values of the constitute

## Undecided Parts

How should we deal with `ActionDiff`?

This PR simply adds a `value` field to `ActionDiff::Pressed`


alice-i-cecile:
> This PR simply adds a value field to ActionDiff::Pressed

I'm fine with this for now, but I'd like to hear what @cBournhonesque would prefer.

cBournhonesque:
I guess my questions is how does `Button + value` differ from `1d Axis`? 
Is it that we want to model the fact that the button can be 'pressed/released', on top of the button?

I wonder if it might not be better to create a new input type for those, since they are a bit niche. I would expect most users to expect a button to just be pressed or not

alice-i-cecile:
> I guess my questions is how does Button + value differ from 1d Axis?

The valid range is [0, 1], not [-1, 1]. It's also used differently semantically.

cBournhonesque:
I see; adding a value works for me!

alice-i-cecile (approved):

PR #654, opened 2024-11-07
Add deserializer function for MouseButton

alradish:
In our project, we want to store and load control to `.ron` file. 
After update to `0.15` there was error `tokens failed to deserialize: no deserialize function was registered for id '"MouseButton"'`
I checked the latest fixes, and it looks like there was just missing `register_buttonlike_input` for `MouseButton`


alice-i-cecile (approved):
Thanks, sorry for missing this!

Issue #646, opened 2024-10-09
WASM CI is making local CI checks harder


alice-i-cecile:
> #637 seems to be preventing me from checking if the CI will pass on my local machine (Win11)

_Originally posted by @Shute052 in https://github.com/Leafwing-Studios/leafwing-input-manager/issues/645#issuecomment-2402914555_

PR #657, opened 2024-11-15
Use native target in ci instead of linux

gak:
* #646

This builds for me in W11 after removing the explicit linux target. Let's see if it will work in GHA.


alice-i-cecile (approved):

alice-i-cecile:
Needs formatting, but this LGTM.

PR #662, opened 2024-12-05
Make the macros deterministic, using BTreeMap instead of HashMap

cormacrelf:
This fixes compile errors in non-cargo build systems, which rely on the rustc crate hash being the same between two compiles (i.e. build with/without codegen).

HashMap order is nondeterministic due to SipHasher initialization, so macro-generated code would change between compiles (in this case, the order of the generated match statement branches).

Can confirm buck2 can now build and link leafwing-input-manager.


alice-i-cecile:
TIL this could be a problem. Sensible fix, thanks.

PR #663, opened 2024-12-06
registries: no need for `static mut` or unsafe + use std's LazyLock

cormacrelf:
This fixes nightly rustc yelling about:

    error: creating a shared reference to mutable static is discouraged
      --> src/user_input/trait_serde.rs:69:37
       |
    69 |         let mut registry = unsafe { AXISLIKE_REGISTRY.write().unwrap() };
       |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^ shared reference to mutable static
       |
       = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html>
       = note: shared references to mutable statics are dangerous; it's undefined behavior if the static is mutated or if a mutable reference is created for it while the shared reference lives

And drops a dependency.

Ref #662 which had build failures.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,6 @@ serde_flexitos = "0.2"
 dyn-clone = "1.0"
 dyn-eq = "0.1"
 dyn-hash = "0.2"
-once_cell = "1.19"
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


alice-i-cecile (approved):
Very pleased to have that gone. Thank you!

PR #664, opened 2024-12-09
Upgrade to Bevy 0.15.0

mnmaita:
Adopts #644

- Branched off [pcwalton:0.15-newer](https://github.com/pcwalton/leafwing-input-manager/tree/0.15-newer) and resolved conflicts with the `main` branch.
- Bumped `bevy_egui` version requirement to the latest compatible version.
- Set Bevy version requirement to 0.15.0.
- Replaced usage of the deprecated `SpriteBundle` with `Sprite` and `Transform` components.
- Applied system ordering suggestions from [this comment](https://github.com/Leafwing-Studios/leafwing-input-manager/pull/644#issuecomment-2527717209).

PR #666, opened 2024-12-10
Fix ignored tests from update to bevy 0.15

CosterM:
Fixes the ignored tests introduced in #664 .

The reason these tests were failing was because the [`find_gamepad`](https://github.com/Leafwing-Studios/leafwing-input-manager/blob/d40715f2cf13aa3d3bcaa8fc5c33ed65c7abb8fd/src/user_input/gamepad.rs#L37) function was changed to rely on the existence of a `Gamepad` component to find a gamepad entity if no entity reference was directly provided, however this was not being added.

### DIFF
--- a/tests/buttonlike.rs
+++ b/tests/buttonlike.rs
@@ -34,7 +34,7 @@ fn test_app() -> App {
     app.insert_resource(input_map)
         .init_resource::<ActionState<TestAction>>();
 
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
Easy! You rock, thanks for tracking this down.

Issue #620, opened 2024-09-09
Inputs serialization


Shatur:
## What problem does this solve?

I would like to be able to serialize inputs to a config file and deserialize it back. But there are a 2 problems:

1. Boxes can't be serialized / deserialized without additional effort.
1. All insertion methods on `InputMap` accept concrete type, not a fat pointer.

## What solution would you like?

Not exactly sure how to solve this...

For 1 we can use something like https://github.com/dtolnay/erased-serde, but it require a separate crate for such a common use case is not very nice. It's also possible to use Bevy's reflection. But it requires manual ser/de implementation for user which is hard to write.
For 2 we can accept `Into<Box<dyn Trait>>` since all methods internally do `Box::new`.

## [Optional] What alternatives have you considered?

But maybe using traits wasn't the right call... Maybe we could have better enums?

## Related work

Caused by https://github.com/Leafwing-Studios/leafwing-input-manager/pull/534.

Shatur:
Looks like the crate already uses https://github.com/Gohla/serde_flexitos, we just need include a usage example for it.

alice-i-cecile:
Adding a generic `insert_boxed` method to `InputMap` is an easy PR, and something I'm firmly in favor of.

Shatur:
Another option would be consider enum dispatch. I.e. keep all the separation like `Buttonlike`, `Axislike`, but make them enums.

PR #669, opened 2024-12-22
Introduce boxed methods for insertion.

CoolSlimbo:
PR introduces additional methods that allow using boxed versions of the traits directly.
Previous methods have been adapted to use the boxed versions underlying, and the boxed methods run all the (previous) logic to them.

Relevant issue: #620

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "leafwing-input-manager"
 description = "A powerful, flexible and ergonomic way to manage action-input keybindings for the Bevy game engine."
-version = "0.16.0"
+version = "0.16.1"
 authors = ["Leafwing Studios"]
 homepage = "https://leafwing-studios.com/"
 repository = "https://github.com/leafwing-studios/leafwing-input-manager"
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,12 @@
 # Release Notes
 
+## Version 0.16.1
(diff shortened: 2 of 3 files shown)


alice-i-cecile (approved):

alice-i-cecile:
Once you add a note to RELEASE.md I'll merge this in :) Thanks!

I think this is a bit clearer / less messy than `impl Trait for Box<dyn Trait>`.

CoolSlimbo:
Finalizing release notes now (suspecting this would be under 0.16.1 as a minor push?), was thinking, should this be listed in the readme as a feature? Especially since it's prime use case is allowing serialized types to be imputed?

alice-i-cecile:
Once we have a full test / example for serialization, yes please to putting it in the README. Until then, we shouldn't make promises that aren't tested :) Please feel free to add anything else required to fix #620: this is a feature I actively want.

CoolSlimbo:
Sorry for the late response, have been busy with life and all.

Besides an example/test (which personally, I'm semi-confused on it due to the ways it can actually be implemented being so vast), that is everything needed for allowing serializing.

At best I might do a test that tests it serde's everything correctly.

alice-i-cecile:
Sounds good. I'm happy to merge this as soon as there's a simple release note entry then. An end-to-end example + integration test would be ideal in the long-run: this is critical functionality to teach and test.

PR #677, opened 2025-01-17
Update bevy_egui to 0.32

vladbat00:
I've disabled default-features (which might save some compilation time for someone), and removed `after(bevy_egui::EguiSet::ProcessInput)` from the system ordering.

Egui writes it to memory whether it wants mouse/keyboard input only during its `end_pass`, which happens in the `PostUpdate` schedule, and since we want to run filtering during `PreUpdate`, the system ordering is redundant (`wants_pointer_input` and others always have a 1-frame delay anyway).


alice-i-cecile:
Looks great :) Add a note to RELEASES.md and I'll merge this in for you.

vladbat00:
Done :)

PR #680, opened 2025-01-31
Fix impossible-to-fulfill requirements for custom input registration api

### LINKED ISSUE #651: Registering custom input kinds seems impossible

Registering custom input kinds with `CentralInputStore` seem impossible. It's documented to do so

```
/// This plugin is added by default by [`InputManagerPlugin`],
/// and will register all of the standard [`UserInput`]s.
///
/// To add more inputs, call [`CentralInputStore::register_input_kind`] during [`App`] setup.

pub struct CentralInputStorePlugin;
```

During `App` setup the `CentralInputStore`, once it's been created, will be inside a resource inside the world inside the app, but `register_input_kind` require a mutable reference to the app in order to add a system.

https://github.com/Leafwing-Studios/leafwing-input-manager/blob/54a361bde6f548e33b4a0491c03d2fa759f5c6cd/src/user_input/updating.rs#L43-L46

This is impossible because you need to hold a mutable reference to the `App` and to the `CentralInputStore` which is inside of it at the same time, exactly a situation the rust borrow checker is designed to prevent.

## Expectation

How do you intuitively expect this to behave?

Instead provide a method that takes the `App`, the control kind, and the `UpdatableInput` type, finds the `CentralInputStore` from the `App`, mutates its state as necessary and then adds the system to the `App` (while no longer holding a mutable reference to the `CentralInputStore` at the same time).

Piefayth:
Fixes #651 by making `register_input_kind` and `register_standard_input_kinds` available on `App` via a new extension trait `InputRegistration`.


alice-i-cecile (approved):

alice-i-cecile:
Nice fix; much appreciated!

PR #685, opened 2025-03-28
update to bevy 0.16

slyedoc:
Straight forward upgrade, I added dependencies for bevy_log, and std for bevy::math and transform, not sure thats ideal.


alice-i-cecile (approved):
Awesome, thank you. I'm going to merge this for now, but we should check with Bushrat on the best way to no_std this crate.

PR #688, opened 2025-04-22
Fix warnings

Koranir:
Based on #686 

Fixes the docs indentation error from clippy, and moves the `#[must_use]` attributes on the `{Button,Axis,DualAxis,TripleAxis}Like` implementations to the trait definition itself to fix a warning on nightly versions.


alice-i-cecile (approved):
Thanks, I really appreciate it :)

PR #693, opened 2025-05-01
Fixes #612 and `previous_duration`

### LINKED ISSUE #612: press_duration example does not work
Compiles and runs without crashing but no effect is found.

Serverator:
Fixes #612 and `previous_duration` not working

`flip()` was called two times, both by update and fixed update, erasing the `previous_duration` value in `ButtonData`'s `Timing`. I switched `flip()` check to only use update data to fix it

### DIFF
--- a/src/action_state/mod.rs
+++ b/src/action_state/mod.rs
@@ -793,7 +793,7 @@ impl<A: Actionlike> ActionState<A> {
         let action_data = self.button_data_mut_or_default(action);
 
         #[cfg(feature = "timing")]
-        if action_data.state.released() {
+        if action_data.update_state.released() {
             action_data.timing.flip();
         }
 
@@ -812,7 +812,7 @@ impl<A: Actionlike> ActionState<A> {
         let action_data = self.button_data_mut_or_default(action);
 
         #[cfg(feature = "timing")]
-        if action_data.state.pressed() {
+        if action_data.update_state.pressed() {
             action_data.timing.flip();
         }
 


mrkiffie:
Hi @Serverator. Thanks for taking a look at this issue. 

I can confirm that the double `flip` call is overriding the previous duration.

## Running the example

```
cargo run --features="timing" --example press_duration
```

Running the `press_duration` example does appear to be working:
> The longer you hold, the faster you dash when released!

https://github.com/user-attachments/assets/8ed82ebb-9219-4070-8d34-0e3195ded646

## Tests

This change however causes the `timing::tests::durations` unit test to fail.

```
cargo test --features=timing timing::tests::durations
```

Are you able to replicate this test failure on your side?

alice-i-cecile:
Oh good spot. I'm happy to merge this once CI is passing: note that the tests themselves may have been buggy!

Serverator:
Test fails as well on my side. Tests use current button state, while duration checks for update state.

I just rechecked all the code and realized that my bandaid fix is a bit flawed, since `update_state` is updated 1 frame later and don't update at all when using `ActionState` manually. 

I see two proper solutions how to fix it properly:
1. Add more `Timing` fields to `ButtonData` — `update_timing` and `fixed_update_timing` and swap them in existing `swap_to_update_state()` and `swap_to_fixed_update_state()`. The only problem I see is that the `Timing` struct is quite byte heavy and would double the size of `ActionKindData` enum, but it would yield proper timings for both update and fixed update schedules.
2. Keep the timing only for update schedule and disable `flip()` check for fixed update schedule. As far as I'm aware, there is no way to check in which schedule `ActionState` is currently is. Maybe adding a field to ActionState with current schedule would be helpful.

~~Or alternately I can keep my current bandaid fix and fix the test so they would update value of `update_state` after setting the data.~~

I'm willing to make a PR for any of the solutions that we agree on.

alice-i-cecile:
I prefer solution 1 here. I think we need to consider how to revise this design in the future (probably in bevy_enhanced_input), but we should get a robust, working, bug-free implementation to compare to.

PR #699, opened 2025-05-15
Added a rebind example.

ForwardFeed:


alice-i-cecile (changes requested):
I really appreciate that this is being done, but I don't think it's complete enough to be helpful. Without serialization, a basic menu or displaying the key that's bound I don't think it's good enough to meaningfully learn from.

ForwardFeed:
Thank you, I'll work on that.

That said, doing a full rebind menu is a lot to explain, 
I was already working on making a Bevy UI tutorial and blending everything into a single example files may take a file of 300-400 lines.

Also If you could help me on formatting the keys for the display? 
Because I have the rest working but since I only manage to retrieve a `Vec<UserInputWrapper>` using the `inputmap.get(action)` which I can boil down to `Box<dyn Buttonlike>` or other input types, but outside of using the Debug trait, I can't find anything out of the box to transform into a displayable string for the UI.

And how can I show serialization, because on my own I use ron, but I can't include it in this project example.

ForwardFeed:
Also, I can't make UI properly work in the examples.

There's no font (and no errors).
Trying to add any Montserrat font in the currently present assets gives this error:
`Failed to load asset 'Montserrat/static/Montserrat-Thin.ttf' with asset loader 'bevy_text::font_loader::FontLoader': unknown magic`.
Adding an external font will work in the context it wants, most of the times it doesn't... I don't want to know.

I was already working on a Bevy UI tutorial on the side, that includes a key rebinder, so I may just give up on this one.
(Please still help me on the key display question)

alice-i-cecile:
> If you could help me on formatting the keys for the display?

Key display is hard: I'm not currently sure of the best solution for that. Winit or the Bevy Discord may have more help.

For fonts, you may need to add features to dev-dependencies.

Don't worry about it; I'm going to close this out for now. We should be focusing our efforts for this upstream in a usage example anyways.

PR #703, opened 2025-07-06
Add public methods to override the fixed_update_state from state

cBournhonesque:
This is useful for networking purposes as what i'm receiving from the remote is a set of Diffs.
Applying the diffs means updating the `state` field, which is usually not enough because the `state` is swapped out for `fixed_update_state` in the FixedUpdateSchedule.

A solution for this would be to apply the diffs AFTER the state has been swapped for fixed_update, but this would mean doing it in RunFixedMainLoop, which is not always feasible for ordering reasons.

Therefore we need a public method to be able to modify directly `fixed_update_state` and not just `state`

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,12 @@
 # Release Notes
 
+## Unreleased
+
+### Usability
+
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
Ugh, CI is red due to the stupid new clippy lint. If you make a PR for that I'll merge and release this faster :p

alice-i-cecile (approved):

PR #705, opened 2025-08-14
fixed doc generation (affects v0.16)

### LINKED ISSUE #668: Docs failed to build for 0.16
## Version

0.16

## Operating system & version

N/A

## What you did

Went to https://docs.rs/leafwing-input-manager/0.16

Documentation

## What actually happened

Got docs.rs error page for doc build of crate

saikuru0:
Fixes #668

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "leafwing-input-manager"
 description = "A powerful, flexible and ergonomic way to manage action-input keybindings for the Bevy game engine."
-version = "0.17.1"
+version = "0.19.0"
 authors = ["Leafwing Studios"]
 homepage = "https://leafwing-studios.com/"
 repository = "https://github.com/leafwing-studios/leafwing-input-manager"
@@ -49,7 +49,7 @@ docs = ['winit/x11', 'winit/wayland']
 
 [dependencies]
 leafwing_input_manager_macros = { path = "macros", version = "0.17" }
-bevy = { version = "0.16.0", default-features = false, features = [
+bevy = { version = "0.17.0", default-features = false, features = [
   "serialize",
   "std",
   "bevy_log",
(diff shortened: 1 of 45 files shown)


alice-i-cecile:
Interesting. I would really prefer not to manually pin winit versions, or specify default features here. This leads to a lot of churn, and can cause problems for users.

Can we pass in settings exclusively for docs.rs somehow?

saikuru0:
hard to do it without any pinning.

pushed a change to give winit's ver some breathing room (any 0.30 patches), disabled defaults, and added x11 and wayland features specifically (thought it's a good idea to do so as a new `docs` feature, but you could move that straight to the winit dep just fine).

Cargo.toml
```diff
@@ -43,12 +43,16 @@ ui = ['bevy/bevy_ui']
 # - Allow 'egui' to take priority over actions when processing inputs.
 egui = ['dep:bevy_egui']
 
+# winit platforms for docs
```
alice-i-cecile, line 46:
```suggestion
# Enable winit platforms to ensure that docs.rs builds correctly
# Users should never need to enable this feature
```

Cargo.toml
```diff
@@ -43,12 +43,16 @@ ui = ['bevy/bevy_ui']
...
+docs = ['winit/x11', 'winit/wayland']
+
 [dependencies]
 leafwing_input_manager_macros = { path = "macros", version = "0.16" }
 bevy = { version = "0.15.0", default-features = false, features = [
   "serialize",
 ] }
 bevy_egui = { version = "0.31", optional = true }
+winit = { version = "0.30", default-features = false }
```
alice-i-cecile, line 55:
```suggestion
# Needed for the `docs` feature. Must be kept in sync with Bevy's version
winit = { version = "0.30", default-features = false, optional = true }
```

alice-i-cecile:
Some suggestions: does that make sense to you?

alice-i-cecile (approved):

alice-i-cecile:
I'll make sure this gets into the next release, thank you!

PR #714, opened 2025-09-21
Fix `ActionState::*_data` returning `Some` for inputs that haven't been updated

### LINKED ISSUE #713: `ActionLike::axis_data` returns `Some` even though `update_axislike` has never been called
## Version

0.17

## Operating system & version

Windows 10

## What you did

I'm implementing a custom `Axislike` for midi controls (raw values are 0-127).

The problem I have is that I don't get midi messages until the midi controls change, so I have no idea what the values are until the user touches the knobs, it's crucial to know whether we have valid axis_data or not for a control, otherwise I would end up up setting volume to zero just because we haven't seen any volume changes yet.

Documentation for `ActionState::axis_data`

```rust
    /// A reference of the [`AxisData`] corresponding to the `action`.
    ///
    /// # Caution
    ///
    /// To access the [`AxisData`] regardless of whether the `action` has been triggered,
    /// use [`unwrap_or_default`](Option::unwrap_or_default) on the returned [`Option`].
    ///
    /// # Returns
    ///
    /// - `Some(AxisData)` if it exists.
    /// - `None` if the `action` has never been triggered (pressed, clicked, etc.).
```

```rust
use bevy::{ecs::system::ResMut, input::InputPlugin, prelude::*};
use leafwing_input_manager::{
    clashing_inputs::BasicInputs,
    prelude::{updating::InputRegistration, *},
    user_input::updating::{CentralInputStore, UpdatableInput},
    InputControlKind,
};
use serde::{Deserialize, Serialize};

#[derive(Resource, Default, Debug)]
pub struct MidiControllerState;

#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, Reflect)]
pub struct MidiAxis(u8);

impl UserInput for MidiAxis {
    fn kind(&self) -> InputControlKind {
        InputControlKind::Axis
    }

    fn decompose(&self) -> BasicInputs {
        BasicInputs::None
    }
}

impl Axislike for MidiAxis {
    fn value(&self, input_store: &CentralInputStore, _entity: Entity) -> f32 {
        // Read the axis value from the central input store
        input_store.value(self)
        // 0.5
    }
}

impl UpdatableInput for MidiAxis {
    type SourceData = Res<'static, MidiControllerState>;

    fn compute(
        mut central_input_store: ResMut<CentralInputStore>,
        _source_data: bevy::ecs::system::StaticSystemParam<Self::SourceData>,
    ) {
        // Update only the second axis input for demonstration
        central_input_store.update_axislike(MidiAxis(1), 0.5);
    }
}

#[derive(Actionlike, Clone, Copy, Debug, PartialEq, Eq, Hash, Reflect)]
enum Action {
    #[actionlike(Axis)]
    Axis0,
    #[actionlike(Axis)]
    Axis1,
    #[actionlike(Axis)]
    Axis2,
}

#[test]
fn axis_data_should_be_none_before_update_axislike() {
    let mut app = App::new();

    app.init_resource::<MidiControllerState>();

    app.add_plugins(MinimalPlugins)
        .add_plugins(InputPlugin)
        .add_plugins(InputManagerPlugin::<Action>::default());

    // Register a new input kind (custom axis input)
    app.register_input_kind::<MidiAxis>(InputControlKind::Axis);

    // Insert a default ActionState resource
    app.init_resource::<ActionState<Action>>();

    // Add an InputMap binding for the custom axis input
    app.insert_resource(
        InputMap::default()
            .with_axis(Action::Axis0, MidiAxis(0))
            .with_axis(Action::Axis1, MidiAxis(1)),
        // Note: Axis2 is not in the input map
    );

    app.update();

    // Before any update, axis_data should be None
    let action_state = app.world().resource::<ActionState<Action>>();

    // not in input map and not updated, should definitely be None
    assert_eq!(action_state.axis_data(&Action::Axis2), None);

    // in input map and updated, should be Some
    assert!(action_state.axis_data(&Action::Axis1).is_some());
    assert_eq!(action_state.axis_data(&Action::Axis1).unwrap().value, 0.5);

    // in input map but not updated, should be None
    assert_eq!(action_state.axis_data(&Action::Axis0), None); // <-- Fails here
}
```
## What actually happened

It seems `ActionData::axis_data` returns `Some` as long as the input is bound, but regardless of whether any data has actually been updated yet.

## Additional information

Any additional information you would like to add such as screenshots, logs, etc.

johanhelsing:
Fixes #713 

This is breaking, as it adds a new method `get_value -> Some<f32>` to `AxisLike`. It also changes the signature of `CentraInputStore::value` to be `Option`

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -4,7 +4,7 @@
 
 ### Breaking Changes (0.19.0)
 
-- The `Buttonlike`, `Axislike` and `DualAxislike` traits now have getters that return `Option`s to indicate whether an input has been seen before. If you implemented your own input types, you should now implement `Buttonlike::get_pressed` instaed of `pressed`, `Axislike::get_value` instead of `value` and `DualAxislike::get_axis_pair` instead of `axis_pair`.
+- The `Buttonlike`, `Axislike`, `DualAxislike` and `TripleAxislike` traits now have getters that return `Option`s to indicate whether an input has been seen before. If you implemented your own input types, you should now implement `Buttonlike::get_pressed` instaed of `pressed`, `Axislike::get_value` instead of `value` and `DualAxislike::get_axis_pair` instead of `axis_pair`.
 
 ### Bugs (0.19.0)
 


alice-i-cecile:
If you add a release note and make sure this is consistent for buttonlike / DualAxisLike I'm happy to merge this.

PR #721, opened 2025-10-26
Add `common_conditions` for `ActionState` components

### LINKED ISSUE #627: Run conditions only work when adding ActionState as resource
## Version

cd50ba147978dfbc4e824fa0aa9695996b2caf13

## Operating system & version

N/A

## What you did

1. Follow [Getting Started](https://github.com/Leafwing-Studios/leafwing-input-manager?tab=readme-ov-file#getting-started).
2. Try to use e.g. `leafwing_input_manager::common_conditions::action_toggle_active` in `run_if`. 

The `run_if` condition should work.

## What actually happened

Observed a panic such as:

```
ERROR app: panicked at /home/teohhanhui/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bevy_ecs-0.14.2/src/system/system_param.rs:534:17:
Resource requested by leafwing_input_manager::common_conditions::action_toggle_active<hexciv::Action>::{{closure}} does not exist: leafwing_input_manager::action_state::ActionState<hexciv::Action>
```

## Additional information

It works if I switch to using a pattern like https://github.com/Leafwing-Studios/leafwing-input-manager/blob/cd50ba147978dfbc4e824fa0aa9695996b2caf13/examples/action_state_resource.rs

it-me-joda:
## What was the problem?

There is an issue described in #627 where `run-if` does not work as expected if the user adds the `ActionState` as a `Component` as described in the Getting Started example.

Closes #627 

## How did you fix it?

This PR would modify the `common_conditions` to also include `ActionState` components.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -14,8 +14,6 @@
 
 ### Usability (0.18.1)
 
-- add the ability to use `register_input_kind` in a `App` builder pattern rather than having to do it separately. [See #706](https://github.com/Leafwing-Studios/leafwing-input-manager/issues/706) for more info
-
 - analog gamepad buttons (triggers) will now report their values between `0.0` and `1.0` and will be considered pressed when above `0.0`
 
 - run conditions provided by `common_conditions` can now find `ActionState`s from either resources or queries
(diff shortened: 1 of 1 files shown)


it-me-joda:
> IMO this behavior should be rolled into the existing common conditions. Check if either of them exist and are pressed.

I have been tinkering with this on and off for a few days. The problem with checking if either exist is Bevy will panic if the Res version isn't there and I didn't find a way around that. I also went down a trait rabbit hole but that's a whole other thing. If you know of a way to accept the Res not existing that would be ideal and yeah I would just add it to the existing implementation. That would be my preferred way of handling it. 

And you'll know better than me on this but to my knowledge eventually Resources will be Components so that may shake this up when that comes but that's a future us problem.

alice-i-cecile (changes requested):
Better! One nit, and then needs a quick release note.

PR #723, opened 2025-10-31
Add a small deadzone for analog inputs

### LINKED ISSUE #722: Analog buttons might want a default deadzone
So while this resolves a lot of bugs it reveals a concern I have when using analog gamepad buttons. Somewhat requently when releasing a a trigger on my controller (DualShock 4 and I don't have any others to test on at the moment) it will report `0.01` every frame until it's touched again which reports as `pressed`. Maybe consider a very small deadzone by default to avoid this scenario. Either way, additional documentation around that scenario would be nice.

_Originally posted by @it-me-joda in https://github.com/Leafwing-Studios/leafwing-input-manager/issues/716#issuecomment-3446643742_

it-me-joda:
## What was the problem?

Occasionally analog inputs would read as low non-zero values when they were not being pressed.

Closes #722

## How did you fix it?

Implement a very small deadzone to account for this issue.

## \[Optional\] What needs special attention?

As discussed in #722, this will be a simple hardcoded deadzone for now. Future work may include making this configurable.

### DIFF
--- a/src/action_state/mod.rs
+++ b/src/action_state/mod.rs
@@ -576,9 +576,10 @@ impl<A: Actionlike> ActionState<A> {
     }
 
     /// Sets the value of the buttonlike `action` to the provided `value`.
-    /// The threshold of `0.02` is to account for a semi-frequent issue
-    /// with analog inputs (e.g., gamepad triggers) reporting very small
-    /// non-zero values when not pressed.
+    /// A threshold of `0.02` must be overcome for the button to count as "pressed"
+    /// This is used to account for a semi-frequent issue with analog inputs
+    /// (e.g., gamepad triggers) reporting very small non-zero values when
+    /// not physically pressed, due to sensor imprecision.
     ///
     /// Also updates the state of the button based on the `value`:
     /// - If `value > 0.02`, the button will be pressed.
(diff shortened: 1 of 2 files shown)


src/action_state/mod.rs
```diff
@@ -576,9 +576,12 @@ impl<A: Actionlike> ActionState<A> {
     }
 
     /// Sets the value of the buttonlike `action` to the provided `value`.
+    /// The threshold of `0.02` is to account for a semi-frequent issue
```
alice-i-cecile:
```suggestion
    /// 
    /// A threshold of `0.02` must be overcome for the button to count as "pressed".
    /// This is used to account for a semi-frequent issue
```

src/action_state/mod.rs
```diff
@@ -576,9 +576,12 @@ impl<A: Actionlike> ActionState<A> {
...
+    /// with analog inputs (e.g., gamepad triggers) reporting very small
+    /// non-zero values when not pressed.
```
alice-i-cecile:
```suggestion
    /// non-zero values when not physically pressed, due to sensor imprecision.
```

alice-i-cecile (changes requested):
A couple of doc suggestions, then this LGTM :)

It might be nice to extract 0.02 out into an in-function constant, but the docs are there at least.

src/action_state/mod.rs
it-me-joda: Updated

PR #729, opened 2025-12-22
bump to 0.18.0-rc.2

it-me-joda:
Bumping to the new RC and bumping the edition to 2024 to match Bevy which triggered some required changes and a bunch of clippy/fmt stuff.

@alice-i-cecile I looked at some old PRs but I'm not entirely sure what else is required for an RC PR but saw there wasn't one and figured I could give it a go taking care of it for you so you have one less thing to worry about.

### DIFF
--- a/tests/gamepad_axis.rs
+++ b/tests/gamepad_axis.rs
@@ -2,7 +2,9 @@
 
 use bevy::ecs::system::SystemState;
 use bevy::input::InputPlugin;
-use bevy::input::gamepad::{GamepadConnection, GamepadConnectionEvent, RawGamepadAxisChangedEvent, RawGamepadEvent};
+use bevy::input::gamepad::{
+    GamepadConnection, GamepadConnectionEvent, RawGamepadAxisChangedEvent, RawGamepadEvent,
+};
 use bevy::prelude::*;
 use leafwing_input_manager::input_processing::{
     WithAxisProcessingPipelineExt, WithDualAxisProcessingPipelineExt,


it-me-joda:
Following the conversation on Discord, I created https://github.com/bevyengine/bevy/issues/22306 and we will not be making changes to the examples at this time.

it-me-joda:
@alice-i-cecile I stole the RELEASES `(Unreleased)` from a previous rc bump PR. I'm not sure what all else to put on that. This is one of my weak points in documentation is knowing what should be in the RELEASES and how detailed it should be as opposed to full documentation.

RELEASES.md
```diff
@@ -1,5 +1,11 @@
 # Release Notes
 
+## Version 0.20.0 (Unreleased)
+
+### Dependencies (0.20.0)
+
+- updated to Bevy 0.18 Release Candidate
```
alice-i-cecile:
```suggestion
- updated to Bevy 0.18.rc.1 (this should be compatible with Bevy 0.18 more broadly, assuming no further relevant breaking changes occur upstream)

```

alice-i-cecile:
> @alice-i-cecile I stole the RELEASES (Unreleased) from a previous rc bump PR. I'm not sure what all else to put on that. This is one of my weak points in documentation is knowing what should be in the RELEASES and how detailed it should be as opposed to full documentation.

I've left a suggestion to improve it :)

In general, the point of the release notes is to give a quick overview of what's changed for people to skim while updating. It shouldn't be the primary source of documentation: instead, think of it like a table of contents that lets users decide when they want to dig deeper.

it-me-joda:
Bevy 0.18.0-rc.2 was just released so I updated to use that instead. This includes the examples window issue we talked about above that turned out to be Wayland specific. I seem to have uncovered another issue. 

[Bevy Cargo Features](https://github.com/bevyengine/bevy/blob/main/docs/cargo_features.md) claims that `bevy_gilrs` enables `gamepad` which is not true as the crate fails to compile without it being explicitly added separately and the [Cargo.toml](https://github.com/bevyengine/bevy/blob/main/Cargo.toml) does not reflect that. I have opened an issue for it https://github.com/bevyengine/bevy/issues/22319

This appears to have been an issue in the first RC as well.

it-me-joda:
We ended up having to use Bevy `gamepad` flag by default to facilitate stuff like:

```rust
#[cfg(not(feature = "gamepad"))]
fn find_gamepad(_: Option<Query<Entity, With<Gamepad>>>) -> Entity {
    Entity::PLACEHOLDER
}
```

While leaving Bevy's `bevy_gilrs` enabled only as needed behind the LWIM `gamepad` flag.

PR #730, opened 2025-12-22
remove ui feature

### LINKED ISSUE #337: Divergent behavior from bevy_input when holding mouse button over egui window

I don't know whether this is intentional or a bug, but I encountered the following scenario:
- I have an action `Foo` that is bound to `MouseInput::Right`
- I draw windows on top of my game using `bevy_editor_pls`
- For my `PrimaryWindow`, I set `cursor.grab_mode = CursorGrabMode::Locked` as well as `cursor.visible = false`
- I do stuff on `ActionState::pressed(Foo)`
- While holding the right mouse button, I "move" my mouse over the `bevy_editor_pls` window. I put move in quotation marks since the cursor is locked and invisible, so I don't actually see anything moving.

LWIM treats this as if I released the right mouse button, reporting `false` on `ActionState::pressed(Foo)` and `true` on `ActionState::just_released(Foo)`. Meanwhile, `Input<MouseButton>::pressed(MouseButton::Right)` stays `true` during the whole process.

## Expectation

Either for LWIM and bevy_input to behave the same way or the divergent behavior to be documented for [`InputKind::Mouse`](https://docs.rs/leafwing-input-manager/0.9.1/leafwing_input_manager/user_input/enum.InputKind.html#variant.Mouse) and/or [`ActionState::pressed`](https://docs.rs/leafwing-input-manager/0.9.1/leafwing_input_manager/action_state/struct.ActionState.html#method.pressed)

### LINKED ISSUE #672: `Res<ButtonInput<MouseButton>>` is cleared when hovering a `bevy_ui` interactable.
## Version

0.16

## Operating system & version

Macos

## What you did

Have an app with `InputManagerPlugin` and a normal `bevy_ui` button.

Hover over the button, left click, and release.

<details>
<summary>Expand Code</summary>

```rust
use bevy::prelude::*;

#[derive(Actionlike, PartialEq, Eq, Clone, Copy, Hash, Debug, Reflect)]
enum Action {
    Run,
    Jump,
}

const NORMAL_BUTTON: Color = Color::srgb(0.15, 0.15, 0.15);

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .add_systems(Update, (button_system, button_input_system))
        .add_plugins(InputManagerPlugin::<Action>::default())
        .run();
}

fn button_system(
    mut interaction_query: Query<(&Interaction, &Children), (Changed<Interaction>, With<Button>)>,
    mut text_query: Query<&mut Text>,
) {
    for (interaction, children) in &mut interaction_query {
        let mut text = text_query.get_mut(children[0]).unwrap();
        match *interaction {
            Interaction::Pressed => {
                **text = "Press".to_string();
            }
            Interaction::Hovered => {
                **text = "Hover".to_string();
            }
            Interaction::None => {
                **text = "Button".to_string();
            }
        }
    }
}

fn button_input_system(input: Res<ButtonInput<MouseButton>>) {
    if input.is_changed() {
        info!("{:?}", input);
    }
}

fn setup(mut commands: Commands) {
    commands.spawn(Camera2d);

    commands
        .spawn(Node {
            width: Val::Percent(100.0),
            height: Val::Percent(100.0),
            align_items: AlignItems::Center,
            justify_content: JustifyContent::Center,
            ..default()
        })
        .with_children(|parent| {
            parent
                .spawn((
                    Button,
                    Node {
                        width: Val::Px(150.0),
                        height: Val::Px(65.0),
                        border: UiRect::all(Val::Px(5.0)),
                        justify_content: JustifyContent::Center,
                        align_items: AlignItems::Center,
                        ..default()
                    },
                    BackgroundColor(NORMAL_BUTTON),
                ))
                .with_child(Text::new("Button"));
        });
}
```
</details>

```rust
Res(ButtonInput { pressed: {}, just_pressed: {}, just_released: {} })
Res(ButtonInput { pressed: {Left}, just_pressed: {Left}, just_released: {} })
Res(ButtonInput { pressed: {}, just_pressed: {}, just_released: {Left} })
```

(This is the output when the button is not hovered)

## What actually happened

1. There is a debug print every frame, because LWIM is triggering `Res<ButtonInput<MouseButton>>`'s change detection.
2. `just_released` and `just_released` are never `Left` at any point.
3. LWIM is seemingly clearing `ButtonInput` on purpose, yet `pressed` occasionally *does* have a value.

```rust
Res(ButtonInput { pressed: {}, just_pressed: {}, just_released: {} })
Res(ButtonInput { pressed: {}, just_pressed: {}, just_released: {} })
Res(ButtonInput { pressed: {}, just_pressed: {}, just_released: {} })
Res(ButtonInput { pressed: {Left}, just_pressed: {}, just_released: {} })
Res(ButtonInput { pressed: {Left}, just_pressed: {}, just_released: {} })
Res(ButtonInput { pressed: {Left}, just_pressed: {}, just_released: {} })
Res(ButtonInput { pressed: {}, just_pressed: {}, just_released: {} }) // IMPORTANT: just_released is empty
Res(ButtonInput { pressed: {}, just_pressed: {}, just_released: {} })
Res(ButtonInput { pressed: {}, just_pressed: {}, just_released: {} })
```

## Additional information

I got here while debugging an app that uses both `leafwing-input-manager` and `bevy-alt-ui-navigation-lite`. `bevy-alt-ui-navigation-lite` uses some convoluted logic and `Res<ButtonInput<MouseButton>>` to determine whether a nav or focus event should be sent when either hovering or clicking. After upgrading to LWIM / Bevy 0.15, this no longer works.

I think that in my case, I can just disable `leafwing-input-manager`'s `ui` feature. But I'm opening this because at the very least, the change detection thing seems like a potential LWIM bug.

it-me-joda:
This is the beginning of removing the UI feature without any sort of documentation or anything around it. I want to land this and the timing feature stuff at the same time since they are both massively breaking things but we also have a Bevy version bump to do first. I think we need to have a new milestone for the version after the new Bevy version bump. 

Closes #672
Closes #337

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -9,7 +9,8 @@
 ### Bugs (0.21.0)
 
 - `ActionDiff`s generated will no longer include entries from disabled `ActionState`s or disabled `Action`s 
-
+- `MouseScroll` and `MouseScrollAxis` now support specifying `MouseScrollUnit::Pixel` or `MouseScrollUnit::Line`
+ 
 ## Version 0.20.0
 
 ### Dependencies (0.20.0)
--- a/src/user_input/mouse.rs
+++ b/src/user_input/mouse.rs
@@ -11,7 +11,7 @@ use bevy::ecs::system::StaticSystemParam;
 use bevy::ecs::system::lifetimeless::SRes;
 use bevy::input::mouse::{
     AccumulatedMouseMotion, AccumulatedMouseScroll, MouseButton, MouseButtonInput, MouseMotion,
-    MouseWheel,
+    MouseScrollUnit, MouseWheel,
 };
 use bevy::input::{ButtonInput, ButtonState};
 use bevy::math::FloatOrd;
@@ -609,6 +609,7 @@ impl Hash for MouseScrollDirection {
 /// ```rust
 /// use bevy::prelude::*;
 /// use bevy::input::InputPlugin;
+/// use bevy::input::mouse::MouseScrollUnit;
 /// use leafwing_input_manager::plugin::CentralInputStorePlugin;
 /// use leafwing_input_manager::prelude::*;
 /// use leafwing_input_manager::user_input::testing_utils::FetchUserInput;
@@ -617,7 +618,12 @@ impl Hash for MouseScrollDirection {
 /// app.add_plugins((InputPlugin, CentralInputStorePlugin));
 ///
 /// // Y-axis movement
-/// let input = MouseScrollAxis::Y;
+/// let mut input = MouseScrollAxis::Y;
+///
+/// // You can configure which [`MouseScrollUnit`] this Axis uses.
+/// // The default is [`MouseScrollUnit::Pixel`]
+/// input.mouse_scroll_unit = MouseScrollUnit::Line;
+/// assert_eq!(input.mouse_scroll_unit, MouseScrollUnit::Line);
 ///
 /// // Scrolling on the chosen axis activates the input
 /// MouseScrollAxis::Y.set_value(app.world_mut(), 1.0);
@@ -634,6 +640,9 @@ pub struct MouseScrollAxis {
     /// The axis that this input tracks.
     pub axis: DualAxisType,
 
+    /// The [`MouseScrollUnit`] this axis uses
+    pub mouse_scroll_unit: MouseScrollUnit,
+
     /// A processing pipeline that handles input values.
     pub processors: Vec<AxisProcessor>,
 }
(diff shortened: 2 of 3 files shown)


alice-i-cecile (changes requested):
Needs a quick note in RELEASES.md please :)

RELEASES.md
```diff
@@ -1,5 +1,11 @@
 # Release Notes
 
+## Version 0.20.0 (Unreleased)
+
+### Breaking Changes (0.20.0)
+
+- the `ui` feature no longer exists
```
alice-i-cecile: This needs more context on what it did, and how users can migrate.

alice-i-cecile: I would briefly summarize what the feature did, and how users can do this on their own. I don't think and example is worth it right now.

it-me-joda: I added a bit more information. Let me know what you think

alice-i-cecile (approved):

alice-i-cecile:
Yes, this will close that issue as well.

PR #741, opened 2026-02-24
Fix Axislike/DualAxislike actions getting stuck

stevesloan:
Fix Axislike and DualAxislike actions getting stuck at stale values when no input is present, most notably after rollback snapshot restoration in networked games.


alice-i-cecile (approved):
Awesome, thanks!

